PR #24, opened 2025-06-06
Feat: txt file input for pageindex

clarenceluo78:
Added support for txt file input

### txt processing features
- **get_txt_page_tokens()**:  
  - Strategy 1: Form Feed Detection - Splits text on `\f` characters (page breaks)
  - Strategy 2: Chapter Detection - (Currently disabled) Detects chapter markers using regex patterns on "CHAPTER" (useful for ebooks)
  - Strategy 3: Character-based Chunking - Fallback method that splits by character count with smart word boundary detection. Defaults to **2000 chars per page**

## Use Case
example result `ebooks_percival_keene_structure.json`, which is generated from http://en.wikipedia.org/wiki/Percival_Keene (part of the NarrativeQA dataset: https://github.com/google-deepmind/narrativeqa)

### DIFF
--- a/docs/ebooks_percival_keene.txt
+++ b/docs/ebooks_percival_keene.txt
@@ -1,0 +1,15667 @@
+The Project Gutenberg eBook of Percival Keene
+    
+This ebook is for the use of anyone anywhere in the United States and
+most other parts of the world at no cost and with almost no restrictions
+whatsoever. You may copy it, give it away or re-use it under the terms
+of the Project Gutenberg License included with this ebook or online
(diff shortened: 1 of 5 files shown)


clarenceluo78:
Modified txt file processing: now supports both token-based and char-based segmentation for txt input

- **Token-based segmentation**: Uses LlamaIndex's TokenTextSplitter for token-level chunks (default)
- **Character-based segmentation**: Traditional approach with word boundary detection as fallback
- **Configurable parameters**: Customize chunk sizes, tokenizers, and overlap settings

PR #29, opened 2025-06-20
Bugfix/loop

littlewwwhite:
Fixed the multi-threaded loop bug and added a quick uv installation method


littlewwwhite:
Support for multiple models has also been added

saccharin98 (changes requested):
Hi, thanks for the contribution! There are a few things that need to be addressed before this can be merged.

**Critical: Needs rebase onto latest `main`**

The async event loop fix in this PR (`asyncio.run(tree_parser(...))` / `asyncio.run(generate_summaries_for_structure(...))`) has already been resolved in `main` via a different approach (`async def page_index_builder()`). Please rebase and align with the current structure to avoid conflicts.

**`sentencepiece` dependency is unused**

The `sentencepiece` package was added to both `requirements.txt` and `pyproject.toml`, but the `sentencepiece_approx` branch in `get_appropriate_tokenizer` only does `len(text) // 4` — `sentencepiece` is never actually imported or used. Please either integrate it properly or remove the dependency entirely, as it adds a heavy C extension with no benefit.

**Minor issues**
- `requirements.txt`: the old `tiktoken==0.7.0` line should be removed rather than commented out
- `generate_toc_continue` and `single_toc_item_index_fixer` lost their default `model` parameter — this is a breaking change and should be mentioned explicitly
- `get_async_openai_client` wraps `openai.AsyncOpenAI` in an `@asynccontextmanager` unnecessarily; the client already supports `async with` directly

PR #65, opened 2026-01-19
fix: prevent infinite loop in extract_toc_content

### LINKED ISSUE #64: fix: prevent infinite loop in extract_toc_content
## Bug Description

The `extract_toc_content` function in `pageindex/page_index.py:183-195` has a broken infinite loop protection mechanism.

## Problem

The while loop exit condition uses `len(chat_history)`, but `chat_history` is rebuilt every iteration:

```python
while not (if_complete == "yes" and finish_reason == "finished"):
    chat_history = [  # Always 2 elements
        {"role": "user", "content": prompt}, 
        {"role": "assistant", "content": response},    
    ]
    # ...
    if len(chat_history) > 5:  # NEVER true!
        raise Exception(...)
```

Since `chat_history` always has exactly 2 elements, `len(chat_history) > 5` is always `False`.

## Impact

If the API never returns `finish_reason == "finished"` or `if_complete == "yes"`, the loop runs forever.

## Fix

Replace with explicit `attempt` counter:

```python
attempt = 0
max_attempts = 5

while not (if_complete == "yes" and finish_reason == "finished"):
    attempt += 1
    if attempt > max_attempts:
        raise Exception('Failed to complete table of contents after maximum retries')
    # ...
```

luojiyin1987:
Fixes #64

## Summary

Fix broken infinite loop protection in `extract_toc_content()`.

## Problem

The function used `len(chat_history) > 5` to exit the loop, but `chat_history` was rebuilt every iteration with exactly 2 elements, making the check never true.

## Change

- Add explicit `attempt` counter initialized to 0
- Add `max_attempts = 5` constant
- Check `attempt > max_attempts` instead of `len(chat_history) > 5`
- Move check to beginning of loop for earlier exit

## Test

Manual code review - the logic now correctly counts iterations and exits after 5 attempts.

### DIFF
--- a/pageindex/page_index.py
+++ b/pageindex/page_index.py
@@ -180,19 +180,22 @@ def extract_toc_content(content, model=None):
     response = response + new_response
     if_complete = check_if_toc_transformation_is_complete(content, response, model)
     
+    attempt = 0
+    max_attempts = 5
+
     while not (if_complete == "yes" and finish_reason == "finished"):
+        attempt += 1
+        if attempt > max_attempts:
+            raise Exception('Failed to complete table of contents after maximum retries')
+
         chat_history = [
-            {"role": "user", "content": prompt}, 
-            {"role": "assistant", "content": response},    
+            {"role": "user", "content": prompt},
+            {"role": "assistant", "content": response},
         ]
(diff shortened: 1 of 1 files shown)


BukeLy (approved):
Thanks for fixing this bug!

PR #72, opened 2026-01-19
Reuse OpenAI clients to improve throughput

### LINKED ISSUE #71: Reuse OpenAI clients to improve throughput
## Problem

Currently, every call to `ChatGPT_API`, `ChatGPT_API_with_finish_reason`, and `ChatGPT_API_async` creates a new OpenAI/AsyncOpenAI client instance. This has significant overhead:

- HTTP connection pool is created on each call
- Reduces throughput in high-concurrency scenarios
- The async version uses `async with` which creates and closes a client on every request

## Solution

- Add module-level singleton clients cached per API key
- Replace client creation with `_get_sync_client`/`_get_async_client` helper functions
- Fix `chat_history` mutation side effect (use list concatenation instead of append)
- Fix error return type inconsistency in `ChatGPT_API_with_finish_reason`
- Add missing `re` import

## Impact

- Improved throughput for batch API calls
- Reduced connection overhead
- Better resource utilization

luojiyin1987:
## Summary

- Add module-level singleton clients cached per API key
- Replace client creation with `_get_sync_client`/`_get_async_client` helper functions
- Fix `chat_history` mutation side effect (use list concatenation instead of append)
- Fix error return type inconsistency in `ChatGPT_API_with_finish_reason`
- Add missing `re` import

Closes #71

### DIFF
--- a/pageindex/utils.py
+++ b/pageindex/utils.py
@@ -2,6 +2,7 @@ import tiktoken
 import openai
 import logging
 import os
+import re
 from datetime import datetime
 import time
 import json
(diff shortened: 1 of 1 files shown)


KylinMountain:
Thanks @luojiyin1987! Since this was opened, the LLM layer was rewritten to route through LiteLLM (#168), so the functions this patches (`ChatGPT_API*` with direct `openai.OpenAI` clients) no longer exist on `main` and the PR now conflicts.

Each concern is already covered in the current code:
- **Client reuse / throughput** (#71): LiteLLM reuses its underlying HTTP connections internally, so there's no per-call client creation anymore.
- **`chat_history` mutation**: `llm_completion` builds messages with `list(chat_history) + [...]`, so the caller's list isn't mutated.
- **Error return consistency**: it returns `"", "error"` on max retries.
- **Missing `re` import**: landed via #281.

Closing as superseded by the LiteLLM migration — thanks for the solid fixes, they're all reflected in `main` now. 🙏

PR #88, opened 2026-01-27
Test: prevent KeyError in physical_index conversion

DankerMu:
## Summary
Add a small regression test to ensure `convert_physical_index_to_int()` handles list items without a `physical_index` key (prevents the KeyError reported in #1).

## Changes
- Add `tests/test_convert_physical_index_to_int.py`

Addresses #1


BukeLy:

PR #89, opened 2026-01-27
new project structure

denis-samatov:


denis-samatov:
Thanks for the detailed feedback and for taking the time to review the PR.

That makes sense — I understand the concern about scope. I combined too many structural and functional changes into a single PR, which made it hard to review safely and incrementally.

You’re also right about the specific blockers:
- translating README.md and CHANGELOG.md into Russian was not appropriate for a public international project;
- the move to the src/ layout is a breaking change and I did not provide a proper migration path;
- some of the added scripts are development-time utilities and should not have been included in the main package PR.

I appreciate the note that parts of the work were still useful, especially the utils modularization and the Pydantic-based config direction.

I’ll take this feedback into account and, if I reopen contributions here, I’ll split them into smaller focused PRs with isolated changes and clearer review boundaries.

Thanks again for the review.

PR #116, opened 2026-02-24
 fix(utils): add missing re import for start_index parsing helpers

GeYugong:
## Summary
  - add missing `import re` in `pageindex/utils.py`
  - fix potential runtime error in:
    - `get_first_start_page_from_text`
    - `get_last_start_page_from_text`

  ## Why
  These helpers call `re.search` / `re.finditer`, but `re` was not imported, which can raise `NameError` when the
  functions are used.

  ## Scope
  Minimal fix only, no behavior change beyond preventing the runtime error.

### DIFF
--- a/pageindex/utils.py
+++ b/pageindex/utils.py
@@ -8,6 +8,7 @@ import json
 import PyPDF2
 import copy
 import asyncio
+import re
 import pymupdf
 from io import BytesIO
 from dotenv import load_dotenv


BukeLy:
### Code review

The fix is correct: `re` is used at lines 276 and 284 (`re.search`, `re.finditer`) but was never imported, so this would cause a `NameError` at runtime. The added `import re` resolves this. The EOF newline fix is also appropriate.

PR #135, opened 2026-03-03
Add ollama support

### LINKED ISSUE #115: [Feature] ollama
Is it possible to use ollama?

rajo69:
Fixes #115

## What this adds

- `--base-url` CLI flag for pointing PageIndex at any OpenAI-compatible endpoint (Ollama, LM Studio, Azure OpenAI, AWS Bedrock, Mistral, etc.).
- `base_url` parameter threaded through all LLM call functions in `page_index.py`, `page_index_md.py`, and `utils.py`.
- `base_url: null` added to `config.yaml` defaults so existing behaviour is completely unchanged when the flag is not provided.
- Fixes a `KeyError` crash in `count_tokens()` where tiktoken could not find an encoding for non-OpenAI model names. Now falls back to `cl100k_base` instead of crashing.

## Usage

```bash
# Ollama
python run_pageindex.py --pdf_path doc.pdf --model llama3.1 --base-url http://localhost:11434/v1

# Any OpenAI-compatible provider
python run_pageindex.py --pdf_path doc.pdf --model mistral --base-url https://your-custom-endpoint/v1
```

## Breaking changes

None. `base_url` defaults to `null` and all existing calls work without any changes.

## Tested

- Default usage (no `--base-url`) - behaviour identical to before
- `count_tokens()` with a non-OpenAI model name - no longer crashes
- `--base-url` flag accepted by CLI and passed through to all LLM calls


ryanwong00:
Since pageIndex used litellm to connect to LLMs, as an interim solution, I set an OS environment variable **OLLAMA_API_BASE** to http://xxxxxx:11434 and set the model to something like **"ollama/...."** and it seems working.

KylinMountain:
Thanks @rajo69! Ollama is now supported out of the box through the LiteLLM integration (#168, merged) — just pass an `ollama/<model>` model name. Closing as covered, appreciate the contribution! 🙏

PR #168, opened 2026-03-16
Integrate litellm for multi-provider LLM support

KylinMountain:
## Summary

- Replace `openai` + `tiktoken` with `litellm` for unified multi-provider LLM support (OpenAI, Azure, Gemini, etc.)
- Rename `ChatGPT_API` → `llm_completion`, `ChatGPT_API_async` → `llm_acompletion` (using `litellm.acompletion` for native async).
- Add backward compatibility: `CHATGPT_API_KEY` auto-aliased to `OPENAI_API_KEY`
- Remove `tiktoken` dependency; use `litellm.token_counter` instead

## Test plan

- [x] Set `OPENAI_API_KEY` and run existing PDF indexing — should behave identically to before
- [x] Set `CHATGPT_API_KEY` (legacy) — should still work via alias
- [x] Try a non-OpenAI provider (e.g. `gemini/gemini-pro`) by setting the appropriate env var

### DIFF
--- a/pageindex/page_index.py
+++ b/pageindex/page_index.py
@@ -300,7 +300,12 @@ def toc_transformer(toc_content, model=None):
         return cleaned_response
     
     last_complete = get_json_content(last_complete)
+    attempt = 0
+    max_attempts = 5
     while not (if_complete == "yes" and finish_reason == "finished"):
+        attempt += 1
+        if attempt > max_attempts:
+            raise Exception('Failed to complete toc transformation after maximum retries')
         position = last_complete.rfind('}')
         if position != -1:
             last_complete = last_complete[:position+2]
--- a/pageindex/utils.py
+++ b/pageindex/utils.py
@@ -74,7 +74,7 @@ async def llm_acompletion(model, prompt):
                 await asyncio.sleep(1)
             else:
                 logging.error('Max retries reached for prompt: ' + prompt)
-                return "Error"
+                return ""
             
             
 def get_json_content(response):
@@ -501,7 +501,7 @@ def remove_structure_text(data):
 def check_token_limit(structure, limit=110000):
     list = structure_to_list(structure)
     for node in list:
-        num_tokens = count_tokens(node['text'], model='gpt-4o')
+        num_tokens = count_tokens(node['text'], model=None)
         if num_tokens > limit:
             print(f"Node ID: {node['node_id']} has {num_tokens} tokens")
             print("Start Index:", node['start_index'])
--- a/run_pageindex.py
+++ b/run_pageindex.py
@@ -3,6 +3,7 @@ import os
 import json
 from pageindex import *
 from pageindex.page_index_md import md_to_tree
+from pageindex.utils import ConfigLoader
 
 if __name__ == "__main__":
     # Set up argument parser
@@ -51,7 +52,6 @@ if __name__ == "__main__":
             raise ValueError(f"PDF file not found: {args.pdf_path}")
             
         # Process PDF file
-        from pageindex.utils import ConfigLoader
         user_opt = {
             'model': args.model,
             'toc_check_page_num': args.toc_check_pages,


KylinMountain:
> ### Code review
> Found 1 issue:
> 
> 1. Two functions retain hard-coded `model="gpt-4o-2024-11-20"` defaults while the rest of the file was updated to `model=None` (which falls back to the configured provider). This defeats the PR's multi-provider goal for these call sites: callers using a non-OpenAI provider will still have OpenAI forced as the model unless they explicitly override it.
> 
> https://github.com/VectifyAI/PageIndex/blob/847782ccb4ae80f5295a872c7198814c5e7b11ea/pageindex/page_index.py#L501-L503
> 
> https://github.com/VectifyAI/PageIndex/blob/847782ccb4ae80f5295a872c7198814c5e7b11ea/pageindex/page_index.py#L734-L736
> 
> 🤖 Generated with [Claude Code](https://claude.ai/code)
> 
> If this code review was useful, please react with 👍. Otherwise, react with 👎.

resolved

KylinMountain:
> ### Code review
> Found 5 issues:
> 
> 1. **`get_page_tokens` retains hardcoded OpenAI model default despite body being updated for multi-provider support.** The PR replaced tiktoken with `litellm.token_counter` inside this function and fixed the same hardcoded-default pattern in `generate_toc_continue` and `single_toc_item_index_fixer` (changing them to `model=None`), but missed `get_page_tokens`. The sole caller at `page_index_main` calls `get_page_tokens(doc)` without passing `model`, so it will always use `"gpt-4o-2024-11-20"` for tokenization regardless of config. (Confidence: 75/100)
> 
> https://github.com/VectifyAI/PageIndex/blob/6bda6e700fa0b35990ef3d44c29ab1cd194bef2c/pageindex/utils.py#L381-L391
> 
> 2. **`openai==1.101.0` remains as a pinned direct dependency in `requirements.txt` despite no longer being directly imported.** The PR removes `import openai` from `utils.py` and replaces all OpenAI calls with litellm, but keeps the pinned openai version. This could cause version conflicts with litellm's own openai dependency. (Confidence: 60/100)
> 
> https://github.com/VectifyAI/PageIndex/blob/6bda6e700fa0b35990ef3d44c29ab1cd194bef2c/requirements.txt#L1-L2
> 
> 3. **`llm_acompletion` returns bare `"Error"` string on max retry exhaustion, inconsistent with `llm_completion` which returns `""`.** PR [fix: make ChatGPT_API_with_finish_reason return consistent tuple #63](https://github.com/VectifyAI/PageIndex/pull/63) specifically fixed this pattern in the sync version (`ChatGPT_API_with_finish_reason`) because callers pass the return value to `extract_json()`. The async variant carries over the old behavior. (Confidence: 40/100 -- pre-existing, not newly introduced)
> 
> https://github.com/VectifyAI/PageIndex/blob/6bda6e700fa0b35990ef3d44c29ab1cd194bef2c/pageindex/utils.py#L58-L77
> 
> 4. **`check_token_limit` hardcodes `model='gpt-4o'`**, which is inconsistent with the PR's multi-provider goal. Not modified by the PR, but `count_tokens` (which it calls) was changed to use `litellm.token_counter`. (Confidence: 30/100 -- pre-existing, on unmodified lines)
> 
> https://github.com/VectifyAI/PageIndex/blob/6bda6e700fa0b35990ef3d44c29ab1cd194bef2c/pageindex/utils.py#L501-L510
> 
> 5. **Unbounded while loop in `toc_transformer`** has no max iteration guard. If `llm_completion` persistently returns `("", "error")`, `finish_reason` will never be `"finished"` and the loop runs forever. The same pattern in `extract_toc_content` was already fixed by commit [ac9ceaf](https://github.com/VectifyAI/PageIndex/commit/ac9ceaf2eece5b41151a332f77fdc523c5dafb63). (Confidence: 30/100 -- pre-existing, PR only changed the function call inside)
> 
> https://github.com/VectifyAI/PageIndex/blob/6bda6e700fa0b35990ef3d44c29ab1cd194bef2c/pageindex/page_index.py#L302-L326
> 
> - If this code review was useful, please react with 👍. Otherwise, react with 👎.

resolved

PR #182, opened 2026-03-26
fix: address Copilot review on utils-split PR

denis-samatov:
## Summary

This PR addresses the automated Copilot review comments left on the `feature/utils-split` PR.

### `pageindex/core/pdf.py`
- **Fix `KeyError` crash**: `get_page_tokens()` now wraps `tiktoken.encoding_for_model()` in `try/except KeyError` and falls back to `cl100k_base` — consistent with the same fallback logic already in `core/llm.py`'s `count_tokens()`

### `tests/conftest.py`
- **Fix broken sys.path**: Replaced `../src` with `..` (project root). The repo has no `src/` directory, so the previous path was silently adding a non-existent location to sys.path, masking potential import failures

### `pageindex/utils.py`
- **Restore backward compatibility**: Added explicit wrappers for the removed legacy symbols `llm_completion` and `llm_acompletion`. Without these, any code importing from `pageindex.utils` would raise `NameError` at runtime. The wrappers delegate to the new `ChatGPT_API` and `ChatGPT_API_async` in `core/llm`, with a thread-executor fallback for the async path

## Testing
All 11 unit tests pass:
```
11 passed in 0.51s
```


JiwaniZakir (commented):
The default values in `config.yaml` and `PageIndexConfig` in `pageindex/config.py` are inconsistent in several places: `toc_check_page_num` is 20 vs. 3, `max_page_num_each_node` is 10 vs. 5, `max_token_num_each_node` is 20000 vs. 4000, and both `if_add_doc_description` and `if_add_node_text` are `false` in the YAML but `True` in the Pydantic model. Since `ConfigLoader` falls back to the Pydantic defaults when no config file is found (e.g., during testing or when the package is installed without the repo's `config.yaml`), this creates two distinct behavioral profiles depending on the deployment context — likely unintentional. The Pydantic field defaults should either match `config.yaml` exactly, or `config.yaml` should be the sole source of truth and the Pydantic defaults should be treated as last-resort fallbacks with that clearly documented.

Additionally, in `ConfigLoader.load`, filtering out `None` values from `user_dict` (`if v is not None`) silently prevents a caller from explicitly clearing a field that has a non-`None` default (e.g., setting `api_key=None` to override a key loaded from the YAML). Consider distinguishing between "not provided" and "explicitly set to None," possibly by using `MISSING` sentinel values or `model_fields_set` from Pydantic v2.

BukeLy:
Thanks for the contribution. Closing this PR because the current packaging, docs, and module architecture have moved forward through later implementations. This branch now targets obsolete structure or content and is no longer suitable to merge as-is.

PR #188, opened 2026-03-27
fix: prevent KeyError crash and context exhaustion in TOC processing

### LINKED ISSUE #163: toc_transformer crashes on large TOCs (~800 page PDF) - JSON truncation + context overflow
## What happened

I ran PageIndex on an 800-page Italian document (a sort of code documentation). The PDF has a valid table of contents with page numbers, and PageIndex detects it correctly. But `toc_transformer` can't handle the size of the TOC.

Two things go wrong:

**1. The LLM runs out of output tokens mid-JSON and the continuation loop starts hallucinating entries.(using `model: "gpt-4o-2024-11-20"`)**

The TOC-to-JSON conversion hits the model's max output length. The code enters the `while` loop in `toc_transformer` that tries to continue generation, but the model starts inventing entries that don't exist in the document. I got entries like this in the output:

```json
{
    "structure": "3173",
    "title": "xyz",
    "page": "3702"
}
```

The document is 800 pages. There is no page 3702.

Eventually the JSON gets cut off mid-string:

And `extract_json` fails with `Expecting value: line 1 column 1 (char 0)`, which then causes a `KeyError: 'completed'` in `check_if_toc_transformation_is_complete` because `extract_json` returned `{}`.

**2. The continuation prompt exceeds gpt-4o's 128k context window.**

The retry loop sends back both the full raw TOC text and the growing (hallucinated) JSON. For a document this size, it eventually tips over:

```
This model's maximum context length is 128000 tokens.
However, your messages resulted in 128017 tokens.
```

The 10-retry loop in `ChatGPT_API` just burns through all retries on the same oversized prompt.

## Full traceback

```
},
{
    "structure": "3176",
    "title": "INDICE DEI RIFERIMENTI DI
ERROR:root:Failed to extract JSON: Expecting value: line 1 column 1 (char 0)
ERROR:root:Failed to parse JSON even after cleanup
Traceback (most recent call last):
  File "PageIndex-main_github_no_changes\run_pageindex.py", line 67, in <module>
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "PageIndex-main_github_no_changes\pageindex\page_index.py", line 1100, in page_index_main
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\utente\AppData\Local\Programs\Python\Python312\Lib\asyncio\runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "C:\Users\utente\AppData\Local\Programs\Python\Python312\Lib\asyncio\base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "PageIndex-main_github_no_changes\pageindex\page_index.py", line 1075, in page_index_builder
    structure = await tree_parser(page_list, opt, doc=doc, logger=logger)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "PageIndex-main_github_no_changes\pageindex\page_index.py", line 1026, in tree_parser
    toc_with_page_number = await meta_processor(
                           ^^^^^^^^^^^^^^^^^^^^^
  File "PageIndex-main_github_no_changes\pageindex\page_index.py", line 956, in meta_processor
    toc_with_page_number = process_toc_with_page_numbers(toc_content, toc_page_list, page_list, toc_check_page_num=opt.toc_check_page_num, model=opt.model, logger=logger)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "PageIndex-main_github_no_changes\pageindex\page_index.py", line 615, in process_toc_with_page_numbers
    toc_with_page_number = toc_transformer(toc_content, model)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "PageIndex-main_github_no_changes\pageindex\page_index.py", line 322, in toc_transformer
    if_complete = check_if_toc_transformation_is_complete(toc_content, last_complete, model)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "PageIndex-main_github_no_changes\pageindex\page_index.py", line 158, in check_if_toc_transformation_is_complete
    return json_content['completed']
           ~~~~~~~~~~~~^^^^^^^^^^^^^
KeyError: 'completed'
```

## Environment
- How I ran the code: python run_pageindex.py --pdf_path xyz.pdf --if-add-node-summary no
- Python 3.12, Windows 10
- pageindex from GitHub (latest main)
- Model: gpt-4o-2024-11-20
- tiktoken: Version: 0.12.0
- Document: ~800 pages, Italian, has a valid TOC with page numbers

Shreyansh1729:
## Summary

Fixes a critical bug where processing large documents (~800+ pages) causes a `KeyError: 'completed'` crash and context window exhaustion during TOC transformation retries.

## Root Cause

Two independent issues combined to break large document processing:

1. **KeyError on truncated LLM output**: When `extract_json()` returns `{}` due to malformed or truncated JSON, direct dictionary access like `json_content['completed']` throws a `KeyError` instead of gracefully handling the failure.

2. **Context window exhaustion in retry loops**: Both `extract_toc_content()` and `toc_transformer()` rebuilt the full prompt context on every retry iteration — embedding the entire raw TOC text AND the growing accumulated response into each new prompt. For 800-page documents, this quickly exceeds the model's 128k token limit.

## Changes Made

- Replace all 5 unsafe `json_content['key']` accesses with `.get('key', default)` to handle empty/malformed LLM responses gracefully
- Refactor `extract_toc_content()` retry loop to grow `chat_history` incrementally — only the delta/new content is appended per iteration, not the full accumulated response
- Refactor `toc_transformer()` retry loop to use `chat_history` instead of re-embedding the entire raw TOC and incomplete JSON directly in each prompt
- Move the initial truncation (`last_complete[:position+2]`) to run once before the retry loop, preventing valid accumulated content from being cut on subsequent passes
- Add explicit `None` guard for `physical_index` before calling `convert_physical_index_to_int`
- Raise explicit `Exception` on max retries so callers fail fast instead of receiving partial results

## Testing

14 mock-based tests added in `tests/test_issue_163.py`:
- 8 tests verify robust key access (empty, malformed, valid responses)
- 4 tests verify `extract_toc_content` retry behavior (first-try success, continuation, max retries raises exception, incremental history growth)
- 2 tests verify `toc_transformer` retry behavior (first-try success, missing key handling)

```
pytest tests/test_issue_163.py -v
======================== 14 passed in 2.80s ========================
```

Closes #163

### DIFF
--- a/pageindex/page_index.py
+++ b/pageindex/page_index.py
@@ -117,9 +117,8 @@ def toc_detector_single_page(content, model=None):
     Please note: abstract,summary, notation list, figure list, table list, etc. are not table of contents."""
 
     response = llm_completion(model=model, prompt=prompt)
-    # print('response', response)
     json_content = extract_json(response)    
-    return json_content['toc_detected']
+    return json_content.get('toc_detected', 'no')
 
 
 def check_if_toc_extraction_is_complete(content, toc, model=None):
@@ -137,7 +136,7 @@ def check_if_toc_extraction_is_complete(content, toc, model=None):
     prompt = prompt + '\n Document:\n' + content + '\n Table of contents:\n' + toc
     response = llm_completion(model=model, prompt=prompt)
     json_content = extract_json(response)
-    return json_content['completed']
+    return json_content.get('completed', 'no')
 
 
 def check_if_toc_transformation_is_complete(content, toc, model=None):
@@ -155,7 +154,7 @@ def check_if_toc_transformation_is_complete(content, toc, model=None):
     prompt = prompt + '\n Raw Table of contents:\n' + content + '\n Cleaned Table of contents:\n' + toc
     response = llm_completion(model=model, prompt=prompt)
     json_content = extract_json(response)
-    return json_content['completed']
+    return json_content.get('completed', 'no')
 
 def extract_toc_content(content, model=None):
     prompt = f"""
@@ -175,27 +174,19 @@ def extract_toc_content(content, model=None):
         {"role": "user", "content": prompt}, 
         {"role": "assistant", "content": response},    
     ]
-    prompt = f"""please continue the generation of table of contents , directly output the remaining part of the structure"""
-    new_response, finish_reason = llm_completion(model=model, prompt=prompt, chat_history=chat_history, return_finish_reason=True)
-    response = response + new_response
-    if_complete = check_if_toc_transformation_is_complete(content, response, model)
+    continue_prompt = "please continue the generation of table of contents, directly output the remaining part of the structure"
     
-    attempt = 0
     max_attempts = 5
-
-    while not (if_complete == "yes" and finish_reason == "finished"):
-        attempt += 1
-        if attempt > max_attempts:
-            raise Exception('Failed to complete table of contents after maximum retries')
-
-        chat_history = [
-            {"role": "user", "content": prompt},
-            {"role": "assistant", "content": response},
-        ]
-        prompt = f"""please continue the generation of table of contents , directly output the remaining part of the structure"""
-        new_response, finish_reason = llm_completion(model=model, prompt=prompt, chat_history=chat_history, return_finish_reason=True)
+    for attempt in range(max_attempts):
+        new_response, finish_reason = llm_completion(model=model, prompt=continue_prompt, chat_history=chat_history, return_finish_reason=True)
         response = response + new_response
+        chat_history.append({"role": "user", "content": continue_prompt})
+        chat_history.append({"role": "assistant", "content": new_response})
         if_complete = check_if_toc_transformation_is_complete(content, response, model)
+        if if_complete == "yes" and finish_reason == "finished":
+            break
+    else:
+        logging.warning('extract_toc_content: max retries reached, returning best effort result')
     
     return response
 
@@ -217,7 +208,7 @@ def detect_page_index(toc_content, model=None):
 
     response = llm_completion(model=model, prompt=prompt)
     json_content = extract_json(response)
-    return json_content['page_index_given_in_toc']
+    return json_content.get('page_index_given_in_toc', 'no')
 
 def toc_extractor(page_list, toc_page_list, model):
     def transform_dots_to_colon(text):
@@ -296,43 +287,40 @@ def toc_transformer(toc_content, model=None):
     if_complete = check_if_toc_transformation_is_complete(toc_content, last_complete, model)
     if if_complete == "yes" and finish_reason == "finished":
         last_complete = extract_json(last_complete)
-        cleaned_response=convert_page_to_int(last_complete['table_of_contents'])
+        cleaned_response = convert_page_to_int(last_complete.get('table_of_contents', []))
         return cleaned_response
     
     last_complete = get_json_content(last_complete)
-    attempt = 0
+    chat_history = [
+        {"role": "user", "content": prompt},
+        {"role": "assistant", "content": last_complete},
+    ]
+    continue_prompt = "Please continue the table of contents JSON structure from where you left off. Directly output only the remaining part."
+
     max_attempts = 5
-    while not (if_complete == "yes" and finish_reason == "finished"):
-        attempt += 1
-        if attempt > max_attempts:
-            raise Exception('Failed to complete toc transformation after maximum retries')
+    for attempt in range(max_attempts):
         position = last_complete.rfind('}')
         if position != -1:
             last_complete = last_complete[:position+2]
-        prompt = f"""
-        Your task is to continue the table of contents json structure, directly output the remaining part of the json structure.
-        The response should be in the following JSON format: 
 
-        The raw table of contents json structure is:
-        {toc_content}
-
-        The incomplete transformed table of contents json structure is:
-        {last_complete}
-
-        Please continue the json structure, directly output the remaining part of the json structure."""
-
-        new_complete, finish_reason = llm_completion(model=model, prompt=prompt, return_finish_reason=True)
+        new_complete, finish_reason = llm_completion(model=model, prompt=continue_prompt, chat_history=chat_history, return_finish_reason=True)
 
         if new_complete.startswith('```json'):
-            new_complete =  get_json_content(new_complete)
-            last_complete = last_complete+new_complete
+            new_complete = get_json_content(new_complete)
+        last_complete = last_complete + new_complete
+
+        chat_history.append({"role": "user", "content": continue_prompt})
+        chat_history.append({"role": "assistant", "content": new_complete})
 
         if_complete = check_if_toc_transformation_is_complete(toc_content, last_complete, model)
-        
+        if if_complete == "yes" and finish_reason == "finished":
+            break
+    else:
+        logging.warning('toc_transformer: max retries reached, returning best effort result')
 
     last_complete = extract_json(last_complete)
 
-    cleaned_response=convert_page_to_int(last_complete['table_of_contents'])
+    cleaned_response = convert_page_to_int(last_complete.get('table_of_contents', []))
     return cleaned_response
     
 
@@ -753,7 +741,7 @@ async def single_toc_item_index_fixer(section_title, content, model=None):
     prompt = toc_extractor_prompt + '\nSection Title:\n' + str(section_title) + '\nDocument pages:\n' + content
     response = await llm_acompletion(model=model, prompt=prompt)
     json_content = extract_json(response)    
-    return convert_physical_index_to_int(json_content['physical_index'])
+    return convert_physical_index_to_int(json_content.get('physical_index'))
 
 
 
--- a/pageindex/page_index.py
+++ b/pageindex/page_index.py
@@ -186,7 +186,7 @@ def extract_toc_content(content, model=None):
         if if_complete == "yes" and finish_reason == "finished":
             break
     else:
-        logging.warning('extract_toc_content: max retries reached, returning best effort result')
+        raise Exception('Failed to complete table of contents extraction after maximum retries')
     
     return response
 
@@ -297,11 +297,12 @@ def toc_transformer(toc_content, model=None):
     ]
     continue_prompt = "Please continue the table of contents JSON structure from where you left off. Directly output only the remaining part."
 
+    position = last_complete.rfind('}')
+    if position != -1:
+        last_complete = last_complete[:position+2]
+
     max_attempts = 5
     for attempt in range(max_attempts):
-        position = last_complete.rfind('}')
-        if position != -1:
-            last_complete = last_complete[:position+2]
 
         new_complete, finish_reason = llm_completion(model=model, prompt=continue_prompt, chat_history=chat_history, return_finish_reason=True)
 
@@ -316,7 +317,7 @@ def toc_transformer(toc_content, model=None):
         if if_complete == "yes" and finish_reason == "finished":
             break
     else:
-        logging.warning('toc_transformer: max retries reached, returning best effort result')
+        raise Exception('Failed to complete TOC transformation after maximum retries')
 
     last_complete = extract_json(last_complete)
 
@@ -741,7 +742,10 @@ async def single_toc_item_index_fixer(section_title, content, model=None):
     prompt = toc_extractor_prompt + '\nSection Title:\n' + str(section_title) + '\nDocument pages:\n' + content
     response = await llm_acompletion(model=model, prompt=prompt)
     json_content = extract_json(response)    
-    return convert_physical_index_to_int(json_content.get('physical_index'))
+    physical_index = json_content.get('physical_index')
+    if physical_index is None:
+        return None
+    return convert_physical_index_to_int(physical_index)
 
 
 
--- a/tests/test_issue_163.py
+++ b/tests/test_issue_163.py
@@ -1,0 +1,135 @@
+import pytest
+import sys
+import os
+from unittest.mock import patch, MagicMock
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
+
+from pageindex.page_index import (
+    check_if_toc_extraction_is_complete,
+    check_if_toc_transformation_is_complete,
+    toc_detector_single_page,
+    detect_page_index,
+    extract_toc_content,
+    toc_transformer,
+)
+
+
+class TestRobustKeyAccess:
+    @patch("pageindex.page_index.llm_completion", return_value="")
+    def test_toc_detector_empty_response(self, mock_llm):
+        result = toc_detector_single_page("some content", model="test")
+        assert result == "no"
+
+    @patch("pageindex.page_index.llm_completion", return_value='{"toc_detected": "yes"}')
+    def test_toc_detector_valid_response(self, mock_llm):
+        result = toc_detector_single_page("some content", model="test")
+        assert result == "yes"
+
+    @patch("pageindex.page_index.llm_completion", return_value="not json at all")
+    def test_toc_detector_malformed_response(self, mock_llm):
+        result = toc_detector_single_page("some content", model="test")
+        assert result == "no"
+
+    @patch("pageindex.page_index.llm_completion", return_value="")
+    def test_extraction_complete_empty_response(self, mock_llm):
+        result = check_if_toc_extraction_is_complete("doc", "toc", model="test")
+        assert result == "no"
+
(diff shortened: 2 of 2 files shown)


JiwaniZakir (commented):
The `single_toc_item_index_fixer` change in `page_index.py` (line 744) swaps a `KeyError` for a potential `TypeError`: `json_content.get('physical_index')` returns `None` when the key is absent, and then `convert_physical_index_to_int(None)` is called. Whether this actually resolves the crash depends entirely on how `convert_physical_index_to_int` handles `None` — worth checking and possibly providing an explicit fallback value (e.g., `.get('physical_index', 0)`) rather than silently passing `None` downstream.

The `for/else` refactor in both `extract_toc_content` and `toc_transformer` is a clear improvement over the manual `attempt` counter, but the `else` branch now only logs a warning instead of raising. This means callers silently receive a partial or potentially malformed result after max retries, which could cause confusing failures further downstream. The original `raise Exception` was at least explicit about the failure; a silent degraded return should at minimum be documented in the function's docstring or validated before returning.

In `toc_transformer`'s loop, `last_complete = last_complete[:position+2]` runs on every iteration — including iterations where `last_complete` already contains successfully appended content from a previous pass. This means valid accumulated output could be silently truncated on each retry cycle if `rfind('}')` finds an interior `}` rather than the terminal one.

Shreyansh1729:
Thanks for the detailed review. All three points are valid — pushed fixes in ce67562:

1. **`physical_index` None guard**: Added an explicit `if physical_index is None: return None` check before calling `convert_physical_index_to_int`, so `None` is never passed downstream.

2. **Silent degradation on max retries**: Restored `raise Exception(...)` in both `extract_toc_content` and `toc_transformer` so callers get an explicit failure instead of a partial result.

3. **Truncation on every iteration**: Moved the `last_complete[:position+2]` logic to run once before the retry loop, so it only trims the initial incomplete response rather than potentially cutting valid accumulated content on subsequent passes.

Shreyansh1729:
@JiwaniZakir check it now

JiwaniZakir:
The `.get('key', default)` changes are the right fix for the KeyError issue. For the context exhaustion fix, it would be worth confirming the refactored retry logic passes only the incremental/delta content on subsequent iterations rather than re-embedding the full accumulated response — that's the critical part for large document handling. The .gitignore discrepancy @KylinMountain flagged should be corrected before merge to keep the description accurate.

Shreyansh1729:
Confirmed on both points:

1. **Incremental chat_history**: The retry loops append only the new delta content per iteration (`chat_history.append({...new_response})`), not the full accumulated response. The LLM sees the conversation history naturally without re-embedding the entire TOC or accumulated JSON in the prompt.

2. **Description updated**: Fixed the PR description to accurately reflect the current behavior — `raise Exception` on max retries, initial truncation moved before the loop, and the `None` guard for `physical_index`.

Shreyansh1729:
@JiwaniZakir it is ready for the merge
@KylinMountain

JiwaniZakir:
The closing rationale is sound — changing the fallback from `{}` to `[]` is the wrong layer to fix this; the real contract violation is that `extract_json()` can return either type depending on the input, which forces every caller to defensively handle both. A cleaner fix would be to make `extract_json()` always return a consistent type (or raise), and fix the 5 unsafe dict accesses with `.get()` as originally proposed without touching the return type. The context exhaustion issue still needs a dedicated fix regardless of how the KeyError is resolved.

Shreyansh1729:
Valid point about the `extract_json()` contract. Changing its return type to always raise on failure (instead of returning `{}`) would be a cleaner long-term fix, but it touches every caller across the codebase — that's a larger refactor.

For this PR, the scope is intentionally narrow: fix the crash from #163 with minimal risk. The `.get()` calls are defensive regardless of how `extract_json()` behaves, and the context exhaustion fix is independent of that.

Happy to open a follow-up PR to normalize `extract_json()` to raise on parse failure if that's preferred over the current silent `{}` return.

JiwaniZakir:
The context exhaustion fix is the more critical of the two — passing the full accumulated response on each retry is quadratic in cost for large documents. To confirm the refactored retry logic is correct, check that the new prompt on iteration N contains only the *remaining* unparsed portion of the TOC, not the concatenation of all prior partial responses. The `.get()` changes are straightforward and correct regardless. Worth noting: if `extract_json()` returns `{}` due to a truncated response mid-retry, the fix prevents the crash but the retry will still re-attempt with stale context unless the prompt construction is also fixed.

Shreyansh1729:
Confirmed — the prompt on iteration N contains only the fixed continuation string, not the accumulated response.

**`extract_toc_content`** (line 181): `llm_completion(prompt=continue_prompt, chat_history=chat_history)` — `continue_prompt` is a static string. `chat_history` grows via `.append(new_response)`, so only the delta from each iteration is added. The full accumulated `response` is never re-embedded in the prompt.

**`toc_transformer`** (line 307): Same pattern — `continue_prompt` is static, `chat_history.append(new_complete)` adds only the new chunk. The original code embedded both `{toc_content}` and `{last_complete}` directly in an f-string prompt every iteration — that was the quadratic cost. Now removed.

Shreyansh1729:
@JiwaniZakir

JiwaniZakir:
The `.get('key', default)` fix for the `KeyError` is straightforward and correct. On the context exhaustion issue, appending only the delta per iteration to `chat_history` is the right approach — but worth verifying the base prompt (with the full TOC text) isn't being reconstructed on each retry call, since that alone can exhaust the window on 800-page docs regardless of how history is managed. If the TOC is being passed as a parameter into the retry loop rather than captured in a closure or outer scope, each iteration may still be re-embedding it.

Shreyansh1729:
The base prompt (containing the full TOC text) is constructed once at line 160-165 and stored in `chat_history[0]` at line 174. It is never reconstructed inside the loop.

The retry loop (line 180-187) only passes `continue_prompt` — a static string with no embedded content. The `chat_history` list grows by appending `{new_response}` (delta only), so the full TOC text exists exactly once in element `[0]` of the history, not re-embedded per iteration.

For `toc_transformer`, same pattern: `toc_content` appears once in `chat_history[0]` (line 295). The old code had an f-string that re-embedded both `{toc_content}` and `{last_complete}` on every retry — that was the quadratic cost eliminated by this refactor.

Shreyansh1729:
@JiwaniZakir

JiwaniZakir:
The incremental `chat_history` approach is the right call — appending only the new delta per iteration keeps the retry cost linear rather than quadratic. One thing worth verifying before merge: confirm the retry loop correctly tracks the *offset* into the raw TOC so each iteration's prompt starts from where the last response left off, not from the beginning of the unparsed block. If that boundary tracking is solid, the context exhaustion fix is sound and this is good to merge. The `extract_json()` contract cleanup @JiwaniZakir raised is worth a follow-up issue, but agreed it doesn't belong in this PR.

Shreyansh1729:
Boundary tracking is handled through the conversation context itself. Here's the exact message sequence the LLM sees on iteration N:

```
[0] user: "Extract TOC from: {full_toc_text}"     ← set once, line 174
[1] assistant: "Section 1... Section 2..."           ← initial partial output
[2] user: "please continue..."                      ← static continuation prompt
[3] assistant: "Section 3... Section 4..."           ← delta from iteration 1
[4] user: "please continue..."
[5] assistant: "Section 5..."                        ← delta from iteration 2
...
[N] user: "please continue..."                      ← current iteration
```

Each assistant message contains only its own chunk. The LLM sees the full conversation thread and naturally continues from where the last response ended. The offset is implicit in the chat sequence — the model knows exactly where it left off because it can see its own prior outputs as separate messages.

`llm_completion` (utils.py line 36) constructs messages as `list(chat_history) + [current_prompt]`, so the full history is preserved in order without re-embedding any content.

Shreyansh1729:
@JiwaniZakir i think this ready now

JiwaniZakir:
The narrow scope is reasonable for an emergency fix, but the `extract_json()` contract issue is worth tracking — returning `{}` on failure silently converts parse errors into missing-key bugs downstream, which is exactly what this PR is patching around. A follow-up that makes `extract_json()` raise (or return `None`) would eliminate this class of bug structurally rather than requiring defensive `.get()` calls at every call site. On the retry logic, one thing to verify: if `chat_history` carries the full accumulated response across iterations and the model's reply on iteration N is also appended before the next call, confirm the total tokens in `chat_history` are being tracked or bounded — otherwise the quadratic growth problem reappears just one layer up.

Shreyansh1729:
The total tokens in `chat_history` do grow with each iteration — that's inherent to any multi-turn continuation since the LLM needs prior context to continue coherently. But the growth is **linear**: each chunk appears exactly once in the history.

The old code was **quadratic** — on iteration N, the prompt f-string embedded the full accumulated output (`{last_complete}`), and then `llm_completion` sent that as a new user message on top of existing history. So the same content was duplicated across layers, and the total tokens scaled as O(N * total_output).

Now it's O(total_output) — each chunk exists once. For an 800-page document, if the TOC output is ~10k tokens split across 5 retries, the final call sees ~10k tokens of history vs the old code seeing ~30k+ due to duplication. Context limits could still be hit on extremely large TOCs, but that would require a fundamentally different approach (chunking the input TOC itself) — a separate concern from this fix.

Agreed on the `extract_json()` follow-up, will open a separate PR for that.

Shreyansh1729:
@JiwaniZakir check it

Shreyansh1729:
@JiwaniZakir , @KylinMountain i think it's ready for merge

Shreyansh1729:
@JiwaniZakir @KylinMountain is there any updates on this one ?

JiwaniZakir:
The `.get()` fixes are straightforward and correct — on all 5 access points. The more important piece is the retry loop refactor; make sure the accumulated response is being passed in a way that doesn't re-embed the full raw TOC text on each iteration, otherwise the fix is incomplete. Also worth confirming the `extract_json()` fallback path actually returns `{}` consistently and not `None`, since `.get()` on `None` will still raise a `TypeError`.

Shreyansh1729:
Confirmed on both points:

1. **extract_json() consistency:** Verified that `extract_json()` in `utils.py` is hardened to return `{}` on any parse error or unexpected exception (lines 127, 130), never `None`. This ensures the `.get()` calls in `page_index.py` are always safe from TypeErrors.

2. **Retry efficiency:** Verified the loop in `extract_toc_content()` (lines 173-187). The `chat_history` is initialized once with the base prompt (containing the TOC text). Inside the loop, we pass only the static `continue_prompt` which contains no embedded text. Only the delta (`new_response`) is appended to history, ensuring the full raw TOC text appears exactly once in the conversation thread, not per iteration.

Shreyansh1729:
@JiwaniZakir it is confirmed and ready for the merge

JiwaniZakir:
The `json_repair` suggestion from @leohuai is worth considering — it would handle partial JSON from truncated LLM outputs more robustly than just falling back to `{}`. That said, the `.get()` fix here is still necessary as a safety net even if we add repair logic, since `json_repair` can return unexpected structures. The bigger win in this PR is the context exhaustion fix; without that, large documents will keep hitting the token limit before we even get a parseable response to repair.

Shreyansh1729:
Thanks for the feedback @JiwaniZakir! Glad we agree on the importance of the context exhaustion fix and the defensive `.get()` calls. The `json_repair` idea is definitely a great candidate for a future refactor to handle truncated outputs even more robustly.

Since the core issues are addressed and the logic is verified, I think this is ready for merge. @KylinMountain

JiwaniZakir:
This PR is still pending review from the maintainers — the fixes are ready and address both the `KeyError` on truncated LLM output and the context window exhaustion on large documents. If @KylinMountain has bandwidth to take a look, the changes are straightforward: `.get()` replacements for the unsafe dict accesses and a refactor to avoid re-embedding the full TOC text on each retry.

Shreyansh1729:
Thanks @JiwaniZakir for the final review. All feedback has been addressed and the 14 unit tests are passing. @KylinMountain whenever you have a chance, this is ready to be merged.

JiwaniZakir:
The verification from @Shreyansh1729 looks solid — the `.get()` changes and `extract_json()` hardening are the right approach. The refactoring of the retry loop to avoid re-embedding the full accumulated response on each iteration is the more critical fix for large documents, so I want to make sure that part of the diff is complete before merging. Can you share the updated `extract_toc_content()` and `toc_transformer()` implementations so I can confirm the context management is handled correctly end-to-end?

Shreyansh1729:
As requested, here is the core logic for `extract_toc_content()` and `toc_transformer()` showing the incremental context management:

### 1. extract_toc_content
```python
# Initialize history with the first turn
chat_history = [
    {"role": "user", "content": prompt}, 
    {"role": "assistant", "content": response},    
]
continue_prompt = "please continue the generation..."

for attempt in range(max_attempts):
    # Static continue_prompt does NOT re-embed the raw {content}
    new_response, _ = llm_completion(model=model, prompt=continue_prompt, chat_history=chat_history, return_finish_reason=True)
    response += new_response
    # Linear O(N) growth: we append only the new delta
    chat_history.append({"role": "user", "content": continue_prompt})
    chat_history.append({"role": "assistant", "content": new_response})
    ...
```

### 2. toc_transformer
```python
# Move truncation logic BEFORE retry loop to preserve JSON structure
position = last_complete.rfind('}')
if position != -1:
    last_complete = last_complete[:position+2]

chat_history = [
    {"role": "user", "content": prompt},
    {"role": "assistant", "content": last_complete},
]
continue_prompt = "Please continue the table of contents..."

for attempt in range(max_attempts):
    # LLM continues from the clean truncation point
    new_complete, _ = llm_completion(model=model, prompt=continue_prompt, chat_history=chat_history, return_finish_reason=True)
    last_complete += new_complete
    # Append only the new result chunk
    chat_history.append({"role": "user", "content": continue_prompt})
    chat_history.append({"role": "assistant", "content": new_complete})
    ...
```

Shreyansh1729:
@JiwaniZakir as requested, I've shared the updated core logic for both functions above. It confirms the linear O(N) context growth and the end-to-end management. Let me know if you need any other details before merge.

JiwaniZakir:
Looks good — the `extract_json()` consistency check is the key safety guarantee here, since the `.get()` calls are only as safe as the return type contract. Approving and merging.

Shreyansh1729:
Thanks for the approval @JiwaniZakir! I see the PR is still 'Open' — let me know if there's any final step I need to take, or if you can merge it now whenever you have a chance. @KylinMountain

JiwaniZakir:
The implementation looks solid — the `.get()` defensive accesses and the context window refactor both address the root causes cleanly. Approving and marking ready to merge.

Shreyansh1729:
@JiwaniZakir any updates ?

Shreyansh1729:
Hi @KylinMountain and @JiwaniZakir, it's been a week since this critical bug fix was approved. Is there anything blocking the merge? I have two other PRs (#216 and #183) waiting as well, and would love to get these finalized. Thank you for your time!

Shreyansh1729:
@JiwaniZakir @KylinMountain please check this

KylinMountain:
Sorry for the long delay here, @Shreyansh1729 — and thanks for the repeated pings and your patience. We've been short on bandwidth working through the backlog, and this one deserved a merge much sooner.

Really appreciate the thorough work: you addressed every review point, added solid tests, and the incremental-`chat_history` refactor is exactly the right fix for the large-document context exhaustion in #163. Merging now. 🙏

(On @JiwaniZakir's good point about the `extract_json()` return-type contract — that broader cleanup is being handled separately in #333, so your narrow, low-risk scope here was the right call.)

Thanks again for pushing this through!

KylinMountain (approved):
Approved — thorough fix for #163 with good test coverage. Thanks for the patience and the pings!

PR #190, opened 2026-03-27
Simplify agentic vectorless RAG demo

rejojer:
## Summary
- Fix bug where `collected.clear()` discarded streamed answer, causing `query_agent()` to always return empty string
- Add `if __name__ == "__main__":` guard so `query_agent` can be imported without side effects
- Remove stale Step 4 from docstring (never implemented)
- Remove misleading `# noqa: F401` on imports that are actually used
- Replace redundant `streamed_this_turn` flag with direct `collected` truthiness check
- Clean up `f"{text}"` → `print(text)`

## Test plan
- [ ] Run `python examples/agentic_vectorless_rag_demo.py` and verify demo completes normally
- [ ] Verify `query_agent()` returns the actual answer string (not empty)
- [ ] Verify `from examples.agentic_vectorless_rag_demo import query_agent` no longer triggers demo execution

### DIFF
--- a/examples/agentic_vectorless_rag_demo.py
+++ b/examples/agentic_vectorless_rag_demo.py
@@ -25,6 +25,7 @@ import requests
 sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 
 from agents import Agent, Runner, function_tool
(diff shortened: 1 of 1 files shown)


rejojer:
### Code review

PR #198, opened 2026-03-29
refactor: fix type errors, restore retrieve module, and use utils namespace in page_index

sahilsasane:
## Summary
  - **Fix basedpyright errors in `client.py`**: replaced missing `PyPDF2` with `pypdf`,
    fixed `str = None` parameters to `str | None = None`, added `assert self.workspace is not None`
    guards in methods that operate on the workspace path, and added `dict[str, Any]` type
    arguments throughout.
  - **Restore `pageindex/retrieve.py`**: the module was deleted but still imported by `client.py`.
    Recreated it with proper type annotations, using `pypdf` instead of `PyPDF2`.
  - **Export `PageIndexClient` from `__init__.py`**: it was defined in `client.py` but not
    re-exported from the package, causing an unknown import symbol error.
  - **Add `print_tree` to `utils.py`**: referenced in examples but missing; added as an alias
    for `print_toc`.
  - **Refactor `page_index.py` imports**: replaced individual function imports from `utils` with
    a single `from . import utils`, and prefixed all call sites with `utils.<fn>` for clarity.
  - **Fix `run_pageindex.py`**: added explicit `from pageindex.page_index import page_index_main`
    since basedpyright cannot resolve names brought in via `*` imports.
  - **Fix `examples/agentic_vectorless_rag_demo.py`**: cast `Path` to `str` where `str | None`
    is expected by `PageIndexClient`.

## Test plan
  - [ ] Run `basedpyright` — no errors in `pageindex/` or `run_pageindex.py`
  - [ ] `uv run run_pageindex.py --pdf_path <file>` executes without import errors
  - [ ] `PageIndexClient` is importable directly from `pageindex`


saccharin98:
Thanks for taking the time to contribute!

However, we're going to close this PR. A general rule in open source is: **one PR, one fix**. This PR bundles several unrelated changes together (dependency swap, type annotations, style reformatting, import restructuring, etc.), which makes it hard to review and risks conflicting with other ongoing contributions.

If you'd like to resubmit, please split these into separate, focused PRs — each addressing a single issue. That makes it much easier for maintainers to review and merge.

Closing for now. Thanks for understanding!

sahilsasane:
Yeah, my bad.
I was only going to fix the type annotations but somehow other commits came into play.

Thank you for taking your time to reply.

PR #200, opened 2026-03-29
Fix AttributeError crashes when LLM returns malformed JSON

### LINKED ISSUE #199: AttributeError crashes in meta_processor and process_no_toc when LLM returns malformed JSON
## Description

When the LLM returns malformed JSON (extra data, empty responses, malformed structure), `extract_json()` in `utils.py` fails and returns `{}` (empty dict) or the downstream functions receive wrong types (str/dict instead of list of dicts). This causes two crashes in the processing pipeline.

## Crash 1: `AttributeError: 'str' object has no attribute 'get'` in `meta_processor`

In `page_index.py`, `meta_processor` runs:

```python
toc_with_page_number = [item for item in toc_with_page_number if item.get('physical_index') is not None]
```

When items in the list are strings instead of dicts (e.g. the LLM returns a list of strings, or `extract_json` parses malformed output into unexpected types), iterating and calling `.get()` fails.

**Full traceback:**

```
File "/app/run_pageindex.py", line 68, in <module>
File "/app/pageindex/page_index.py", line 1098, in page_index_main
...
File "/app/pageindex/page_index.py", line 958, in meta_processor
    toc_with_page_number = [item for item in toc_with_page_number if item.get('physical_index') is not None]
AttributeError: 'str' object has no attribute 'get'
```

## Crash 2: `AttributeError: 'dict' object has no attribute 'extend'` in `process_no_toc`

In `page_index.py`, `process_no_toc` runs:

```python
toc_with_page_number = generate_toc_init(group_texts[0], model)
for group_text in group_texts[1:]:
    toc_with_page_number_additional = generate_toc_continue(toc_with_page_number, group_text, model)
    toc_with_page_number.extend(toc_with_page_number_additional)
```

When `generate_toc_init()` returns a dict instead of a list (because `extract_json()` returns `{}` on parse failure), calling `.extend()` on a dict raises AttributeError.

**Full traceback:**

```
File "/app/run_pageindex.py", line 68, in <module>
File "/app/pageindex/page_index.py", line 1098, in page_index_main
...
File "/app/pageindex/page_index.py", line 577, in process_no_toc
    toc_with_page_number.extend(toc_with_page_number_additional)
```

## Root Cause

`extract_json()` in `utils.py` returns `{}` (empty dict) on JSON parse failure. All callers expect a list of dicts. The fallback return value should be `[]` (empty list), and callers should validate the return type.

## Reproduction

- Use a local vLLM endpoint with smaller models (e.g. not GPT-4/Claude-class)
- Process Office documents converted to PDF via LibreOffice
- The model frequently returns malformed JSON responses, triggering these crashes
- All 7 test documents failed with these errors

## Environment

- Python 3.11
- Local vLLM endpoint
- Office documents converted to PDF via LibreOffice

martindecz:
## Summary

Fixes two `AttributeError` crashes that occur when the LLM returns malformed JSON output, causing `extract_json()` to fail and downstream functions to receive wrong types.

Closes #199

## Changes

1. **`pageindex/utils.py`** - `extract_json()` now returns `[]` (empty list) instead of `{}` (empty dict) on parse failure. All callers expect a list of dicts, so the fallback value should match that contract.

2. **`pageindex/page_index.py`** - `process_no_toc()` now validates that `generate_toc_init()` and `generate_toc_continue()` return lists before calling `.extend()`. If they return a non-list type (e.g. dict from failed JSON parsing), the code logs a warning and either resets to an empty list or skips the result.

3. **`pageindex/page_index.py`** - `meta_processor()` list comprehension now filters out non-dict items with `isinstance(item, dict)` before calling `.get()`, preventing `'str' object has no attribute 'get'` errors.

## Reproduction context

These crashes are reliably triggered when using a local vLLM endpoint with smaller models that frequently produce malformed JSON (extra data, empty responses, malformed structure). All 7 test documents (Office docs converted to PDF) failed with these errors.

## Test plan

- [ ] Process PDFs with a model that produces clean JSON output (no regression)
- [ ] Process PDFs with a model that produces malformed JSON output (crashes prevented, graceful degradation)

### DIFF
--- a/pageindex/page_index.py
+++ b/pageindex/page_index.py
@@ -584,8 +584,14 @@ def process_no_toc(page_list, start_index=1, model=None, logger=None):
     logger.info(f'len(group_texts): {len(group_texts)}')
 
     toc_with_page_number= generate_toc_init(group_texts[0], model)
+    if not isinstance(toc_with_page_number, list):
+        logger.warning(f'generate_toc_init returned {type(toc_with_page_number).__name__} instead of list, resetting to empty list')
+        toc_with_page_number = []
     for group_text in group_texts[1:]:
-        toc_with_page_number_additional = generate_toc_continue(toc_with_page_number, group_text, model)    
+        toc_with_page_number_additional = generate_toc_continue(toc_with_page_number, group_text, model)
+        if not isinstance(toc_with_page_number_additional, list):
+            logger.warning(f'generate_toc_continue returned {type(toc_with_page_number_additional).__name__} instead of list, skipping')
+            continue
         toc_with_page_number.extend(toc_with_page_number_additional)
     logger.info(f'generate_toc: {toc_with_page_number}')
 
@@ -967,7 +973,7 @@ async def meta_processor(page_list, mode=None, toc_content=None, toc_page_list=N
     else:
         toc_with_page_number = process_no_toc(page_list, start_index=start_index, model=opt.model, logger=logger)
             
-    toc_with_page_number = [item for item in toc_with_page_number if item.get('physical_index') is not None] 
+    toc_with_page_number = [item for item in toc_with_page_number if isinstance(item, dict) and item.get('physical_index') is not None]
     
     toc_with_page_number = validate_and_truncate_physical_indices(
         toc_with_page_number, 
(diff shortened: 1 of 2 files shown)


martindecz:
Closing this PR — our type guards were based on incorrect assumptions about the logger interface (JsonLogger doesn't have .warning()) and the fallback return type change from {} to [] broke more call sites than it fixed (6 out of 13 callers expect dict). The root cause is better addressed in extract_json() itself with robust JSON repair, not in individual callers. Apologies for the noise.

PR #205, opened 2026-03-31
feat: add uv dependency management with pyproject.toml

littlewwwhite:
## Summary

- Add `pyproject.toml` with PEP 621 project metadata and pinned dependencies aligned with existing `requirements.txt`
- Add `uv.lock` for reproducible dependency resolution
- Add `.python-version` (3.12) for uv python version pinning
- Update README with `uv sync` as recommended install method, keeping `pip` as fallback

## Motivation

The project currently only uses a flat `requirements.txt` for dependency management. Adding [uv](https://docs.astral.sh/uv/) support provides:

- **Reproducible builds** — `uv.lock` locks the full dependency tree, preventing implicit version drift
- **Faster installs** — uv is 10-100x faster than pip
- **Standard metadata** — `pyproject.toml` follows PEP 517/518/621, the modern Python packaging standard
- **Backward compatible** — existing `pip install -r requirements.txt` workflow is preserved

## Test plan

- [x] `uv sync` installs all dependencies successfully
- [x] Core imports verified (`litellm`, `pymupdf`, `PyPDF2`, `dotenv`, `yaml`)
- [x] `uv.lock` uses official PyPI registry (no mirror URLs)
- [ ] Maintainer review of Python version constraint (`>=3.10`) and dependency pins


BukeLy:
Thanks for the contribution. Closing this PR because the current packaging, docs, and module architecture have moved forward through later implementations. This branch now targets obsolete structure or content and is no longer suitable to merge as-is.

PR #206, opened 2026-03-31
Add llm_kwargs support for local LLM backends (e.g. Ollama)

omarsherif0:
### Adds `llm_kwargs` support to enable local LLM backends like Ollama 
without needing to set environment variables

## Changes
#add `llm_kwargs` to `PageIndexClient`
#pass `llm_kwargs` through PDF and Markdown indexing paths
#inject `llm_kwargs` into `litellm.completion()` and `litellm.acompletion()`
#add `llm_kwargs: {}` to config defaults

### DIFF
--- a/pageindex/client.py
+++ b/pageindex/client.py
@@ -32,7 +32,7 @@ class PageIndexClient:
 
     For agent-based QA, see examples/agentic_vectorless_rag_demo.py.
     """
-    def __init__(self, api_key: str = None, model: str = None, retrieve_model: str = None, workspace: str = None):
+    def __init__(self, api_key: str = None, model: str = None, retrieve_model: str = None, workspace: str = None, llm_kwargs: dict | None = None):
         if api_key:
             os.environ["OPENAI_API_KEY"] = api_key
         elif not os.getenv("OPENAI_API_KEY") and os.getenv("CHATGPT_API_KEY"):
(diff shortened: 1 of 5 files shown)


KylinMountain:
Thanks @omarsherif0 — this fills a real gap: `llm_completion` currently hardcodes `temperature=0` and passes nothing else, so there's no way to set e.g. Ollama's `num_ctx`, `max_tokens`, or a per-call `api_base`. The `ContextVar` + `llm_kwargs_scope` propagation is a sensible way to avoid threading the param through the whole call stack, and it's async-safe. A few things before merge:

**1. (bug) `temperature` collision.** `litellm.completion(..., temperature=0, **llm_kwargs)` raises `TypeError: got multiple values for keyword argument 'temperature'` as soon as a user puts `temperature` in `llm_kwargs` — one of the most common overrides. Please merge into one dict so kwargs can override the default:
```python
params = {"temperature": 0, **llm_kwargs}
response = litellm.completion(model=model, messages=messages, **params)
```

**2. Scope only covers indexing.** The scope is set in `page_index_main` / `md_to_tree`, so retrieval/QA `llm_completion` calls outside it won't pick up `llm_kwargs`. If that's intended, a docstring note would help; otherwise the retrieval path needs the same wrapping.

**3. (minor) Diff noise.** Wrapping the whole body of `page_index_main`/`md_to_tree` in the `with` re-indents everything, making the change hard to review — `token = _DEFAULT_LLM_KWARGS.set(...)` + `finally: reset(token)` avoids that if you prefer.

**Nice to have:** a short test (e.g. `llm_kwargs={'num_ctx': 8192}` reaches the call) and a one-line docs example for local models. FYI the failing `Analyze (python)` check is a repo-side ruleset issue, not your PR.

BukeLy:

PR #228, opened 2026-04-11
feat: Universal Local LLM & Custom Endpoint Support (Ollama, LM Studio, vLLM, etc.)

sreenivas134:
This PR introduces comprehensive support for local LLMs and custom API endpoints. While existing proposals often focus on specific providers (like Ollama), this implementation leverages LiteLLM's full capabilities to provide a universal, config-driven interface that works with any OpenAI-compatible or local provider (LM Studio, vLLM, LocalAI, Ollama, etc.) with minimal configuration.

### ✨ Key Features
- **Universal api_base Support:** Adds a global api_base attribute that can be configured via config.yaml, environment variables, or directly through the CLI.
- **Automated Provider Normalization:** Implements intelligent model-name routing. If a custom endpoint is detected (like LM Studio at localhost:1234), the system automatically ensures the correct openai/ prefix is applied, removing the need for users to manually prefix their models for most local inference servers.
- **Robust Direct Parameter Passing:** Unlike implementations that rely solely on environment variables, this PR passes the api_base directly to LiteLLM's completion calls, ensuring reliability across multi-provider sessions and preventing environment pollution.
- **Centralized Configuration:** All LLM setup logic is consolidated into a new setup_litellm helper in utils.py, making the codebase cleaner and easier to maintain.

###  Verification
- Tested with LM Studio and Ollama local endpoints.
- Verified environment variable isolation (no cross-contamination between different base_url settings).
- Confirmed recursive support for both PDF and Markdown processing flows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,23 @@ Create a `.env` file in the root directory with your LLM API key, with multi-LLM
 OPENAI_API_KEY=your_openai_key_here
 ```
 
-### 3. Generate PageIndex structure for your PDF
+### 3. Local LLM Support (Optional)
+
+PageIndex supports local LLMs via LM Studio, Ollama, etc. You can specify a custom `api_base` in your `.env` or pass it as an argument.
+
+**Example for LM Studio:**
+```bash
+# In .env
+# OPENAI_API_BASE=http://localhost:1234/v1
+```
+
+**Example for Ollama:**
+```bash
+# In .env
+# OLLAMA_API_BASE=http://localhost:11434
+```
+
+### 4. Generate PageIndex structure for your PDF
 
 ```bash
 python3 run_pageindex.py --pdf_path /path/to/your/document.pdf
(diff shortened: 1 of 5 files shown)


saccharin98:
Thanks for the contribution.

  We are going to close this as redundant with LiteLLM's existing functionality. PageIndex already routes LLM calls through LiteLLM, and LiteLLM already
  supports provider selection via model strings like `openai/...`, `anthropic/...`, `gemini/...`, `ollama/...`, etc., as well as custom endpoints via its built-
  in `api_base` parameter.

  We do not want to duplicate LiteLLM provider/endpoint routing inside PageIndex with ad-hoc CLI flags, environment-variable mutation, global state, or model-
  name rewriting. That would create inconsistent behavior across the CLI, SDK, PDF, and Markdown paths.

  A future contribution would be more useful if it documents the recommended LiteLLM usage, or adds a minimal PageIndex config pass-through for `api_base` that
  is scoped, tested, and free of global side effects.

PR #229, opened 2026-04-12
fix: add --base_url CLI option for custom LLM endpoints

### LINKED ISSUE #224: Base url missing
I noted that there r request to add in the base url to allow hosted e.g ollama to be performed . However the changes are not reflected in the main branch . Can advise ? Thanks

octo-patch:
Fixes #224

## Problem

Users running local LLM servers (Ollama, vLLM, LM Studio, etc.) have no way to specify a custom API base URL through the CLI. The `OPENAI_API_BASE` environment variable works when set manually, but there is no `--base_url` argument, making the workflow unnecessarily awkward.

## Solution

- **`run_pageindex.py`**: Added `--base_url` argument that sets `os.environ["OPENAI_API_BASE"]` before any LLM calls are made.
- **`pageindex/utils.py`**: Updated `llm_completion` and `llm_acompletion` to read `OPENAI_API_BASE` from the environment and forward it to litellm as `api_base`, so all LLM calls in the pipeline respect the custom endpoint.

The change is fully backward-compatible: if `--base_url` is not provided and `OPENAI_API_BASE` is not set, behaviour is identical to before.

## Usage

```bash
# Ollama
python run_pageindex.py --pdf_path doc.pdf \
  --model ollama/llama3 \
  --base_url http://localhost:11434

# vLLM / LM Studio or any OpenAI-compatible server
python run_pageindex.py --pdf_path doc.pdf \
  --model openai/my-model \
  --base_url http://localhost:8000/v1
```

## Testing

Verified that the argument is parsed correctly and `OPENAI_API_BASE` is set before any LLM call. The `api_base` kwarg is only injected into litellm when the env var is present, so there is no overhead for standard OpenAI usage.

### DIFF
--- a/pageindex/utils.py
+++ b/pageindex/utils.py
@@ -34,12 +34,14 @@ def llm_completion(model, prompt, chat_history=None, return_finish_reason=False)
         model = model.removeprefix("litellm/")
     max_retries = 10
     messages = list(chat_history) + [{"role": "user", "content": prompt}] if chat_history else [{"role": "user", "content": prompt}]
+    api_base = os.getenv("OPENAI_API_BASE")
     for i in range(max_retries):
         try:
             response = litellm.completion(
                 model=model,
                 messages=messages,
                 temperature=0,
+                **({"api_base": api_base} if api_base else {}),
             )
             content = response.choices[0].message.content
             if return_finish_reason:
@@ -64,12 +66,14 @@ async def llm_acompletion(model, prompt):
         model = model.removeprefix("litellm/")
     max_retries = 10
     messages = [{"role": "user", "content": prompt}]
+    api_base = os.getenv("OPENAI_API_BASE")
     for i in range(max_retries):
         try:
             response = await litellm.acompletion(
                 model=model,
                 messages=messages,
                 temperature=0,
+                **({"api_base": api_base} if api_base else {}),
             )
             return response.choices[0].message.content
         except Exception as e:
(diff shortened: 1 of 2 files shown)


saccharin98:
Thanks for the contribution.

  I think we should be careful about the scope here. If the goal is to let users choose an LLM provider, this is already supported through LiteLLM model strings.
  `run_pageindex.py` already accepts `--model`, and the LLM helpers pass that model through to `litellm.completion()` / `litellm.acompletion()`, so users can select providers
  with values like `anthropic/...`, `gemini/...`, `ollama/...`, `openai/...`, etc. In that sense, adding another provider-selection path would be redundant.

  For issue #224 specifically, the missing piece is not provider selection, but a custom OpenAI-compatible base URL. That is a valid feature request, but I do not think this PR
  is ready as-is:

  1. The new `--base_url` only applies to the `run_pageindex.py` CLI path. It does not cover the Python SDK / `PageIndexClient` local-mode path, so the configuration surface
  would be inconsistent.
  2. Reading `OPENAI_API_BASE` inside every `llm_completion()` / `llm_acompletion()` call and unconditionally forwarding it as `api_base` affects all LiteLLM providers, not just
  OpenAI-compatible endpoints. That can be surprising when users are using non-OpenAI providers via LiteLLM.
  3. There are no tests for the new behavior. At minimum, we should mock `litellm.completion` and `litellm.acompletion` and verify that `api_base` is omitted by default and
  passed only when explicitly configured.

  I would suggest either closing this as redundant if the intent is provider selection, or reworking it as a proper custom endpoint feature wired through the shared config/API
  layer rather than only through a CLI env-var side effect.

PR #246, opened 2026-04-24
fix: make markdown parser robust

### LINKED ISSUE #245: Markdown parser fails on common edge cases
## Description

The current `MarkdownParser` in `pageindex/parser/markdown.py` has several robustness issues. It silently drops content or produces incorrect results on valid Markdown inputs that are common in real-world documents.

**All cases below have been verified against the old parser — 6 out of 7 confirmed failing.**

### 1. Content before the first header is silently dropped

~~~markdown
This is an introduction paragraph.
Some important context here.

# Chapter 1
Body text.
~~~

**Expected:** The preamble text should be captured as a node.
**Actual:** `_build_nodes` only iterates over headers, so everything before `# Chapter 1` is lost. (1 node returned, preamble missing)

### 2. Headerless Markdown files produce zero nodes

~~~markdown
Just some plain text document
with multiple lines of content.
No headers at all.
~~~

**Expected:** At least one node containing the full content.
**Actual:** `_extract_headers` returns `[]`, `_build_nodes` returns `[]` — the entire document is silently discarded.

### 3. YAML frontmatter — not a separate bug

~~~markdown
---
title: My Document
author: Alice
---

# Introduction
Hello world.
~~~

**Verified:** The frontmatter does NOT leak into node content — it is silently dropped as part of the preamble (same root cause as Case 1). However, the new parser now properly extracts frontmatter into `metadata`, which is the correct behavior.

### 4. Setext-style headers are not recognized

~~~markdown
Main Title
==========

Some content here.

Sub Section
-----------

More content.
~~~

**Expected:** Two nodes — H1 "Main Title" and H2 "Sub Section".
**Actual:** Zero headers detected, zero nodes returned. Entire document lost.

### 5. Tilde code fences are not tracked

~~~~markdown
# Header

~~~python
# This is a comment, not a header
def foo():
    pass
~~~

# Real Header
~~~~

**Expected:** 2 nodes: "Header" and "Real Header".
**Actual:** 3 nodes — the comment `# This is a comment, not a header` inside the tilde fence is incorrectly detected as a header.

### 6. Mismatched fence lengths cause incorrect header detection

Example input (a 4-backtick fence containing 3-backtick lines):

    # Before

    ````python
    # Inside 4-backtick fence
    ```
    # Still inside — 3 backticks can't close a 4-backtick fence
    ```
    ````

    # After

**Expected:** 2 headers: "Before" and "After".
**Actual:** 3 headers detected: "Before", "Still inside", "After". The parser uses a simple toggle on any `` ``` `` line, so it incorrectly "closes" the 4-backtick fence at the first 3-backtick line.

### 7. ATX closing hashes are included in title

~~~markdown
## My Section ##

Content here.
~~~

**Expected:** Title is `"My Section"`.
**Actual:** Title is `"My Section ##"` — the trailing `##` is not stripped.

## Reproduction Script

Self-contained script that inlines the old parser and runs all 7 cases. Save as `test_old_parser.py` and run with `python test_old_parser.py`.

<details>
<summary>Click to expand test script</summary>

~~~~python
"""Test the OLD markdown parser against the 7 edge cases from issue #245."""
import re
import sys
import tempfile
from pathlib import Path
from dataclasses import dataclass


# --- Inline the old parser so we don't need to mess with imports ---

@dataclass
class ContentNode:
    content: str
    tokens: int
    title: str | None = None
    index: int | None = None
    level: int | None = None
    images: list | None = None

@dataclass
class ParsedDocument:
    doc_name: str
    nodes: list
    metadata: dict | None = None

def count_tokens(text, model=None):
    return len(text.split())  # cheap approximation for testing

class OldMarkdownParser:
    """Exact copy of the parser BEFORE the fix (commit 6d29886)."""
    def supported_extensions(self):
        return [".md", ".markdown"]

    def parse(self, file_path, **kwargs):
        path = Path(file_path)
        model = kwargs.get("model")
        with open(path, "r", encoding="utf-8") as f:
            content = f.read()
        lines = content.split("\n")
        headers = self._extract_headers(lines)
        nodes = self._build_nodes(headers, lines, model)
        return ParsedDocument(doc_name=path.stem, nodes=nodes)

    def _extract_headers(self, lines):
        header_pattern = r"^(#{1,6})\s+(.+)$"
        code_block_pattern = r"^```"
        headers = []
        in_code_block = False
        for line_num, line in enumerate(lines, 1):
            stripped = line.strip()
            if re.match(code_block_pattern, stripped):
                in_code_block = not in_code_block
                continue
            if not in_code_block and stripped:
                match = re.match(header_pattern, stripped)
                if match:
                    headers.append({
                        "title": match.group(2).strip(),
                        "level": len(match.group(1)),
                        "line_num": line_num,
                    })
        return headers

    def _build_nodes(self, headers, lines, model):
        nodes = []
        for i, header in enumerate(headers):
            start = header["line_num"] - 1
            end = headers[i + 1]["line_num"] - 1 if i + 1 < len(headers) else len(lines)
            text = "\n".join(lines[start:end]).strip()
            tokens = count_tokens(text, model=model)
            nodes.append(ContentNode(
                content=text,
                tokens=tokens,
                title=header["title"],
                index=header["line_num"],
                level=header["level"],
            ))
        return nodes


# --- Helper ---
def parse_string(parser, content, name="test"):
    with tempfile.NamedTemporaryFile(mode="w", suffix=".md", delete=False) as f:
        f.write(content)
        f.flush()
        return parser.parse(f.name)


# --- Tests ---
parser = OldMarkdownParser()
failures = []
passes = []

# Case 1: Preamble before first header
result = parse_string(parser, """This is an introduction paragraph.
Some important context here.

# Chapter 1
Body text.
""")
preamble_found = any("introduction" in n.content for n in result.nodes)
if preamble_found:
    passes.append("Case 1: Preamble")
else:
    failures.append(
        "Case 1: Preamble before first header -> DROPPED "
        "(got {} nodes, none contain 'introduction')".format(len(result.nodes))
    )

# Case 2: Headerless file
result = parse_string(parser, """Just some plain text document
with multiple lines of content.
No headers at all.
""")
if len(result.nodes) > 0:
    passes.append("Case 2: Headerless file")
else:
    failures.append("Case 2: Headerless file -> 0 nodes returned, entire document LOST")

# Case 3: YAML frontmatter
result = parse_string(parser, """---
title: My Document
author: Alice
tags: [python, markdown]
---

# Introduction
Hello world.
""")
frontmatter_in_content = any("title: My Document" in n.content for n in result.nodes)
if not frontmatter_in_content:
    passes.append("Case 3: Frontmatter (not a separate bug, dropped with preamble)")
else:
    failures.append("Case 3: YAML frontmatter -> leaked into node content")

# Case 4: Setext headers
result = parse_string(parser, """Main Title
==========

Some content here.

Sub Section
-----------

More content.
""")
titles = [n.title for n in result.nodes if n.title]
if "Main Title" in titles and "Sub Section" in titles:
    passes.append("Case 4: Setext headers")
else:
    failures.append(
        "Case 4: Setext headers -> not recognized "
        "(got titles: {}, {} nodes)".format(titles, len(result.nodes))
    )

# Case 5: Tilde fences
result = parse_string(parser, """# Header

~~~python
# This is a comment, not a header
def foo():
    pass
~~~

# Real Header
""")
titles = [n.title for n in result.nodes if n.title]
if "This is a comment, not a header" not in titles:
    passes.append("Case 5: Tilde fences")
else:
    failures.append(
        "Case 5: Tilde fences -> comment inside ~~~ detected as header "
        "(got titles: {})".format(titles)
    )

# Case 6: Mismatched fence lengths
content = """# Before

````python
# Inside 4-backtick fence
```
# Still inside
```
````

# After
"""
result = parse_string(parser, content)
titles = [n.title for n in result.nodes if n.title]
has_fake = "Inside 4-backtick fence" in titles or "Still inside" in titles
if not has_fake:
    passes.append("Case 6: Mismatched fence lengths")
else:
    failures.append(
        "Case 6: Mismatched fence lengths -> fake headers detected "
        "inside code block (got titles: {})".format(titles)
    )

# Case 7: ATX closing hashes
result = parse_string(parser, """## My Section ##

Content here.
""")
if result.nodes and result.nodes[0].title == "My Section":
    passes.append("Case 7: ATX closing hashes")
else:
    actual_title = result.nodes[0].title if result.nodes else "N/A"
    failures.append(
        "Case 7: ATX closing hashes -> title is '{}' "
        "instead of 'My Section'".format(actual_title)
    )

# --- Report ---
print("=" * 60)
print("OLD PARSER EDGE CASE VERIFICATION")
print("=" * 60)
print()
for p in passes:
    print(f"  PASS  {p}")
for f in failures:
    print(f"  FAIL  {f}")
print()
print(f"Result: {len(failures)} failed, {len(passes)} passed out of 7")
sys.exit(1 if failures else 0)
~~~~

</details>

**Output:**

~~~
============================================================
OLD PARSER EDGE CASE VERIFICATION
============================================================

  PASS  Case 3: Frontmatter (not a separate bug, dropped with preamble)
  FAIL  Case 1: Preamble before first header -> DROPPED (got 1 nodes, none contain 'introduction')
  FAIL  Case 2: Headerless file -> 0 nodes returned, entire document LOST
  FAIL  Case 4: Setext headers -> not recognized (got titles: [], 0 nodes)
  FAIL  Case 5: Tilde fences -> comment inside ~~~ detected as header (got titles: ['Header', 'This is a comment, not a header', 'Real Header'])
  FAIL  Case 6: Mismatched fence lengths -> fake headers detected inside code block (got titles: ['Before', 'Still inside', 'After'])
  FAIL  Case 7: ATX closing hashes -> title is 'My Section ##' instead of 'My Section'

Result: 6 failed, 1 passed out of 7
~~~

## Impact

These issues affect real-world Markdown documents:
- README files and documentation commonly have preamble text before the first header
- Technical docs use setext headers and various code fence styles
- Some Markdown editors add closing hashes to ATX headers
- A headerless `.md` file being completely discarded is a data loss bug

## Environment

- `pageindex` version: 0.3.0.dev1
- Python: 3.11

### PR DESCRIPTION
**Pull request by saccharin98**
This PR closes #245

### DIFF
--- a/pageindex/parser/markdown.py
+++ b/pageindex/parser/markdown.py
@@ -71,10 +71,12 @@ class MarkdownParser:
         fence_pattern = None  # tracks the char and min length to close
 
         for line_num, line in enumerate(lines, 1):
-            stripped = line.strip()
+            indent = self._indent_width(line)
+            content = line.lstrip(" \t")
+            stripped = content.strip()
 
             # Track fenced code blocks
-            fence_match = _FENCE_OPEN.match(stripped)
+            fence_match = _FENCE_OPEN.match(content) if indent <= 3 else None
             if fence_match:
                 marker = fence_match.group(1)
                 if not in_fence:
@@ -91,6 +93,10 @@ class MarkdownParser:
             if in_fence:
                 continue
 
+            # Four-space/tab indented lines are code blocks, not headers.
+            if indent >= 4:
+                continue
+
             # ATX headers: # Title
             atx = _ATX_HEADER.match(stripped)
             if atx:
@@ -104,8 +110,9 @@ class MarkdownParser:
             # Setext headers: underline on next line detected by looking back
             # We check if *this* line is an underline and the previous line is text
             if line_num >= 2:
-                prev = lines[line_num - 2].strip()  # previous line (0-indexed)
-                if self._is_setext_paragraph_candidate(prev):
+                prev_line = lines[line_num - 2]
+                prev = prev_line.strip()  # previous line (0-indexed)
+                if self._indent_width(prev_line) < 4 and self._is_setext_paragraph_candidate(prev):
                     if _SETEXT_H1.match(stripped):
                         headers.append({
                             "title": prev,
@@ -122,6 +129,18 @@ class MarkdownParser:
                         continue
 
         return headers
+
+    @staticmethod
+    def _indent_width(line: str) -> int:
+        width = 0
+        for char in line:
+            if char == " ":
+                width += 1
+            elif char == "\t":
+                width += 4 - (width % 4)
+            else:
+                break
+        return width
 
     @staticmethod
     def _is_closing_fence(line: str, fence_pattern: tuple[str, int] | None) -> bool:
--- a/pageindex/parser/markdown.py
+++ b/pageindex/parser/markdown.py
@@ -1,5 +1,7 @@
 import re
 from pathlib import Path
+import yaml
+
 from .protocol import ContentNode, ParsedDocument
 from ..index.utils import count_tokens
 
@@ -9,6 +11,9 @@ _SETEXT_H1 = re.compile(r"^={3,}\s*$")
 _SETEXT_H2 = re.compile(r"^-{3,}\s*$")
 _FENCE_OPEN = re.compile(r"^(`{3,}|~{3,})")
 _FRONTMATTER_FENCE = re.compile(r"^---\s*$")
+_BLOCKQUOTE = re.compile(r"^>")
+_LIST_ITEM = re.compile(r"^(?:[-+*]|\d{1,9}[.)])(?:\s+|$)")
+_TABLE_ROW = re.compile(r"^\|.*\|$|^[^|]+\|[^|]+$")
 
 
 class MarkdownParser:
@@ -23,9 +28,9 @@ class MarkdownParser:
             content = f.read()
 
         lines = content.split("\n")
-        lines, metadata = self._strip_frontmatter(lines)
+        lines, metadata, line_offset = self._strip_frontmatter(lines)
         headers = self._extract_headers(lines)
-        nodes = self._build_nodes(headers, lines, model)
+        nodes = self._build_nodes(headers, lines, model, line_offset)
 
         return ParsedDocument(
             doc_name=path.stem,
@@ -34,22 +39,29 @@ class MarkdownParser:
         )
 
     @staticmethod
-    def _strip_frontmatter(lines: list[str]) -> tuple[list[str], dict | None]:
+    def _strip_frontmatter(lines: list[str]) -> tuple[list[str], dict | None, int]:
         """Strip YAML frontmatter (--- delimited) from the beginning of the file.
 
-        Returns the remaining lines and raw frontmatter as metadata.
+        Returns the remaining lines, raw frontmatter metadata, and removed line count.
         """
         if not lines or not _FRONTMATTER_FENCE.match(lines[0]):
-            return lines, None
+            return lines, None, 0
 
         for i in range(1, len(lines)):
             if _FRONTMATTER_FENCE.match(lines[i]):
                 raw = "\n".join(lines[1:i])
+                try:
+                    parsed = yaml.safe_load(raw)
+                except yaml.YAMLError:
+                    return lines, None, 0
+                if not isinstance(parsed, dict):
+                    return lines, None, 0
+
                 remaining = lines[i + 1:]
-                return remaining, {"frontmatter": raw}
+                return remaining, {"frontmatter": raw}, i + 1
 
         # No closing fence found — not valid frontmatter, return as-is
-        return lines, None
+        return lines, None, 0
 
     def _extract_headers(self, lines: list[str]) -> list[dict]:
         """Extract all ATX and setext headers, respecting fenced code blocks."""
@@ -92,7 +104,7 @@ class MarkdownParser:
             # We check if *this* line is an underline and the previous line is text
             if line_num >= 2:
                 prev = lines[line_num - 2].strip()  # previous line (0-indexed)
-                if prev and not _FENCE_OPEN.match(prev) and not _ATX_HEADER.match(prev):
+                if self._is_setext_paragraph_candidate(prev):
                     if _SETEXT_H1.match(stripped):
                         headers.append({
                             "title": prev,
@@ -110,8 +122,21 @@ class MarkdownParser:
 
         return headers
 
+    @staticmethod
+    def _is_setext_paragraph_candidate(line: str) -> bool:
+        return bool(line
+                    and not _FENCE_OPEN.match(line)
+                    and not _ATX_HEADER.match(line)
+                    and not _BLOCKQUOTE.match(line)
+                    and not _LIST_ITEM.match(line)
+                    and not _TABLE_ROW.match(line))
+
     def _build_nodes(
-        self, headers: list[dict], lines: list[str], model: str | None
+        self,
+        headers: list[dict],
+        lines: list[str],
+        model: str | None,
+        line_offset: int = 0,
     ) -> list[ContentNode]:
         if not headers:
             # No headers — entire content becomes a single node
@@ -119,7 +144,8 @@ class MarkdownParser:
             if not text:
                 return []
             tokens = count_tokens(text, model=model)
-            return [ContentNode(content=text, tokens=tokens, index=1)]
+            index = self._first_content_line(lines, line_offset)
+            return [ContentNode(content=text, tokens=tokens, index=index)]
 
         nodes = []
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by saccharin98** [comment]
Review notes after testing the branch locally (`69 passed, 2 skipped`):

The switch to `markdown-it-py` fixes the main CommonMark cases from #245: preamble/headerless docs, setext headings, tilde/long fences, and ATX closing hashes all look covered by tests. I found two follow-up issues worth fixing before merge:

1. `pageindex/parser/markdown.py:27` strips frontmatter before extracting headings, but the parser never carries the removed line offset forward. That changes `ContentNode.index` / structure `line_num` from original-file line numbers to line numbers in the stripped buffer. For example, a file with 3 lines of frontmatter, a blank line, then `# Intro` now returns `index=2` instead of the original line 5. The old parser exposed original Markdown line numbers, and downstream storage uses `n.index` as the page/line identifier, so this is a user-visible regression. Please preserve the frontmatter offset and add it back when assigning heading indexes.

2. `_strip_frontmatter()` treats any file that starts with `---` and later has another `---` line as YAML frontmatter. That can silently delete valid Markdown content/thematic-break based documents, e.g. `---\n# Intended heading\n---\n...` becomes metadata and the heading disappears. Since `pyyaml` is already a dependency, it would be safer to validate the fenced block as frontmatter and add a regression test for a real Markdown document that starts with a thematic break.

Once those two are addressed, the overall direction looks good to me.

**Comment by rejojer** [comment]
### Code review

Found 2 issues:

1. `_strip_frontmatter` returns `lines[i + 1:]` without carrying the stripped-line offset forward. `_extract_headers` then enumerates the trimmed buffer starting at `line_num=1`, and `_build_nodes` assigns that to `ContentNode.index`. So for a file with N lines of frontmatter, every downstream `node.index` is shifted by N relative to the original file. This is user-visible: `pageindex/index/pipeline.py` `build_tree_from_levels` maps `node.index` to `"line_num"`, and `get_md_page_content` uses `node.index` for range-based content lookup — both now return positions in the stripped buffer, not the real file. (The PR author already flagged this same point in an earlier self-review note.)

https://github.com/VectifyAI/PageIndex/blob/50aab66c0dab242274cce0f54c1f9e6345d0e9cb/pageindex/parser/markdown.py#L36-L52

Suggested fix: have `_strip_frontmatter` also return the stripped line count, and add it to each `header["line_num"]` (or to `ContentNode.index`) before returning.

2. Setext header detection accepts any non-empty, non-fence, non-ATX previous line as a valid title. It does not reject list items (`- item`, `* item`, `1. item`), blockquotes (`> ...`), or table rows. Per CommonMark, setext headers require a *paragraph* preceding the underline. So a list like `- item\n---\n` will be promoted to an H2 header with title `- item`, silently reshaping the document tree.

https://github.com/VectifyAI/PageIndex/blob/50aab66c0dab242274cce0f54c1f9e6345d0e9cb/pageindex/parser/markdown.py#L91-L109

Suggested fix: additionally reject `prev` lines that start with list markers (`-`, `*`, `+`, `<digits>.`, `<digits>)`) or blockquote (`>`), or verify `prev` looks like a paragraph.

**Comment by saccharin98** [comment]
Review notes after testing the latest commit (`a349946`) locally: `75 passed, 2 skipped`.

The prior line-offset and setext non-paragraph issues look addressed. I found two remaining Markdown parsing gaps worth fixing before merge:

1. `pageindex/parser/markdown.py:76-88` treats any line starting with the fence marker as a closing fence while already inside a fence. In CommonMark, a closing fence cannot have non-space trailing characters. Repro:

~~~markdown
# Before

```
```not a closing fence
# Still code
```

# After
~~~

The parser currently returns titles `Before` and `Still code`; `After` is swallowed into the fake section. This is close to the original fence robustness issue in #245, just with an info-string-like line inside the code block. Please require the closing fence line to be only the marker plus spaces/tabs, and add this regression test.

2. `pageindex/parser/markdown.py:10-11` only recognizes setext underlines with three or more `=`/`-` characters. Valid Markdown allows one or more marker characters, so `Title\n-\n` and `Title\n=\n` currently become a headerless content node instead of an H2/H1 section. Please relax those regexes and cover the one-character cases.

The rest of the PR direction looks good.

saccharin98:
Follow-up review after the latest commit (`c244249`): local tests pass with `80 passed, 2 skipped`.

No blocking issues found in the latest changes. I rechecked the indented-code and fence cases:

- 4-space/tab-indented `# ...` lines are now ignored as code.
- 3-space-indented headings still parse.
- 3-space-indented fences still suppress fake headings inside.
- Fence closing with trailing spaces works.
- Fence-like lines with trailing text stay inside the code block.

One low-priority compatibility gap remains: valid empty ATX headings like `#` or `##` are still treated as plain/headerless content instead of heading nodes. That is a CommonMark edge case, but I would not consider it blocking for this PR.

BukeLy:

PR #281, opened 2026-05-18
Adds missing re import

Mouli04:


KylinMountain (approved):
Thanks @Mouli04 — verified the missing import (both helpers use re but it was never imported). Merging! 🙏

PR #282, opened 2026-05-19
docs: document additional run_pageindex CLI options

MackDing:
Updates the README CLI reference so it includes the additional `run_pageindex.py` options that are already supported by the script, including markdown thinning and output-detail flags.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,13 +173,17 @@ python3 run_pageindex.py --pdf_path /path/to/your/document.pdf
 You can customize the processing with additional optional arguments:
 
 ```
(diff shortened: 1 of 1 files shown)


MackDing:
Quick follow-up: this PR appears mergeable from my side; the remaining blocker looks like required maintainer review. Happy to adjust if you prefer a different approach.

MackDing:
Checking in on this PR. It is still open and appears to be waiting on maintainer review.

I am happy to adjust the scope or add verification details if that would make review easier.

Issue #286, opened 2026-05-21
Installation - Requirements with version of litellm need python-dotenv==1.0.1 but conflict with requirements python-dotenv==1.2.2


e68026063-oss:
As the title sugggests, when installing from requirements using uv, there is a conflict between litellm need and what the requirements fixed for the version of python-dotenv.

`
× No solution found when resolving dependencies:
  ╰─▶ Because litellm==1.83.7 depends on python-dotenv==1.0.1 and you require litellm==1.83.7, we can conclude that you require python-dotenv==1.0.1.
      And because you require python-dotenv==1.2.2, we can conclude that your requirements are unsatisfiable.
`

yerlikayao:
`sed -i 's/python-dotenv==1.2.2/python-dotenv==1.0.1/' requirements.txt` work s for me

PR #291, opened 2026-05-23
Fix: python-dotenv version conflict with litellm (#286)

50RISHU:
Fixed python-dotenv version that conflict with litellm while installing from requirements.txt using uv.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,5 +2,5 @@ litellm==1.83.7
 # openai-agents  # optional: required for examples/agentic_vectorless_rag_demo.py
 pymupdf==1.26.4
 PyPDF2==3.0.1
-python-dotenv==1.2.2
+python-dotenv==1.0.1
 pyyaml==6.0.2


requirements.txt
```diff
@@ -2,5 +2,5 @@ litellm==1.83.7
 # openai-agents  # optional: required for examples/agentic_vectorless_rag_demo.py
 pymupdf==1.26.4
 PyPDF2==3.0.1
-python-dotenv==1.2.2
+python-dotenv==1.0.1
```
50RISHU, line 5: Confirmed — litellm has a known incompatibility with python-dotenv>=1.2. Downgrading to 1.0.1 is the right fix. Tested locally, no issues.

KylinMountain (approved):
Thanks @50RISHU! Verified — litellm 1.83.7 requires exactly python-dotenv==1.0.1, so the current 1.2.2 breaks a fresh pip/uv install. This fixes it. Merging, closes #286. 🙏

KylinMountain:
Thanks @50RISHU — you correctly spotted the install conflict! The catch is python-dotenv==1.0.1 has a known moderate vuln (GHSA-mf9w-mj56-hr94, fixed in 1.2.2), so pinning down to it trips Dependency Review. The cleaner fix is to bump litellm to ≥1.84 (which relaxed its python-dotenv==1.0.1 pin) and keep python-dotenv==1.2.2. Closing in favor of that, but really appreciate you surfacing #286! 🙏

Issue #296, opened 2026-05-25
The bug of reading contents from md file in demo code


HaiyangPeng:
After running the code PageIndex/examples/agentic_vectorless_rag_demo.py, the following case will occur:
1. when the file is md, LLM will infer multiple page_num, e.g., 53,57 and the _get_md_page_content may return contents from line 53 to line 57. This is not correct for md file, so I fix this bug by changing the agent prompt to use ONLY the exact line_num values from the
structure, limiting the tool description of get_page_content and using precise line extraction in _get_md_page_content.

I will provide a PR for fixing this bug

BukeLy:
Thanks for the report. Closing because the affected legacy JSON/Markdown/TOC path has since been replaced or covered by later parser validation, cleanup, retry, and error-handling changes. If this still reproduces on the latest release, please open a new issue with the input shape, model/provider, package version, and a minimal traceback.

PR #297, opened 2026-05-25
fix the bug of md content extraction.

HaiyangPeng:
This PR aims to fix the bug of md content extraction #296 . @BukeLy @KylinMountain

### DIFF
--- a/examples/agentic_vectorless_rag_demo.py
+++ b/examples/agentic_vectorless_rag_demo.py
@@ -12,7 +12,7 @@ Agent tools:
   - get_page_content()       — retrieve text content of specific pages
 
 Steps:
-  1 — Index a PDF and view its tree structure index
+  1 — Index a PDF/md and view its tree structure index
   2 — View document metadata
   3 — Ask a question (agent reasons over the index and auto-calls tools)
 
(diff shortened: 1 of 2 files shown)


KylinMountain:
I don’t think we should let the LLM know whether the file is PDF or Markdown, just to decide which method to call — that logic should be handled by the code, not the prompt.

Also, if we later add support for other formats like PPT, would we need to update the prompt again?

Separately, I’m still unclear about the root cause of the issue. Is it:

Failing to extract the correct line-numbered content?

A prompt design issue?

Or a function naming problem?

HaiyangPeng:
> I don’t think we should let the LLM know whether the file is PDF or Markdown, just to decide which method to call — that logic should be handled by the code, not the prompt.
> 
> Also, if we later add support for other formats like PPT, would we need to update the prompt again?
> 
> Separately, I’m still unclear about the root cause of the issue. Is it:
> 
> Failing to extract the correct line-numbered content?
> 
> A prompt design issue?
> 
> Or a function naming problem?

Yes, the prompt can be relatively free and there is no need to limit the file format. But if we add this limitation for LLM, it can stably infer correct line numbers. For example, in my test case, LLM infers “37-39, 4-5“ line ranges of a markdown file, which will output [4, 5, 37, 38, 39] after `_parse_pages`. The core error will occur in `min_line, max_line = min(page_nums), max(page_nums)`, because all the line contents between min_line and max_line, i.e., 4, 39 will be extracted, which is incorrect. Therefore, I replace `min_line, max_line = min(page_nums), max(page_nums)` with if` ln and ln in requested and ln not in seen:` in `_get_md_page_content` to avoid this error.
In addition, I constrain the LLM to avoid infer ranges by the following prompt:
`- For Markdown documents: use ONLY the exact line_num values from the
structure,
separated by commas. Example: "3,53,69". Do NOT use ranges.`   @KylinMountain

BukeLy:

PR #302, opened 2026-05-26
feat(filesystem): add PageIndex FileSystem and PIFS CLI

BukeLy:
## Summary

This PR adds PageIndex FileSystem (PIFS): a filesystem-like interaction system for agents working inside a PageIndex workspace, plus a `pifs` CLI and an `ask` / `chat` loop built on the same command surface.

The core purpose is to help an agent quickly locate the right file in a workspace, then combine that filesystem context with PageIndex structure, metadata, and projection indexes to retrieve precise file evidence.

## Goal

PIFS gives agents a stable filesystem-like interface to PageIndex workspaces:

- inspect workspace shape with familiar folder commands;
- locate relevant files inside an explicit path scope;
- use PageIndex-backed structure, metadata, and semantic projections for precise evidence retrieval;
- keep reads bounded, auditable, and tied to concrete virtual file paths.

`browse` is part of the file-location step. It ranks file candidates within a folder scope when folder names and exact filters are not enough. Evidence still comes from bounded `cat`, `grep`, and PageIndex structural reads.

## What Changed

- Added the PIFS core model: virtual folders, registered files, metadata status, PageIndex/projection status, path resolution, and SQLite persistence.
- Added `pifs`, a shell-style CLI for workspace navigation, file discovery, metadata filtering, source reads, imports, and agent execution.
- Added PageIndex-backed registration for PDF, Markdown, and text files, including structural reads, generated metadata, and summary projection indexing.
- Added `pifs add` for atomic local imports into workspace-owned artifacts.
- Added `pifs ask` and `pifs chat`, where the agent uses the same read-only filesystem commands available to users.
- Added PIFS Semantic Folder as an explicit build step for flat or weakly organized corpora: `pifs semantic-folder build [source_scope]` materializes a generated `<source_scope>/semantic` tree from canonicalized `domain` / `topic` metadata.
- Added command guardrails for bounded reads, lexical `grep -R`, path ambiguity, projection dimension mismatches, atomic import cleanup, and semantic-folder rebuild safety.
- Expanded regression coverage across storage, command parsing/rendering, registration, add rollback, browse behavior, structural reads, metadata generation, semantic indexes, and semantic-folder materialization.

## Command Surface

- Global flags: `--workspace`, `--env-file`, `--json`
- Workspace defaults: `pifs set workspace <path>`
- Navigation and inspection: `pifs ls`, `pifs tree`, `pifs find`, `pifs stat`
- File discovery: `pifs browse [-R] <folder> "<query>" [--space summary|entity|relation] [--where JSON] [--page N]`
- Evidence reads: `pifs cat <path> --structure|--page|--range|--all`, `pifs grep [-R] <pattern> <path>`
- Imports and generated views: `pifs add <physical_path> <virtual_path>`, `pifs semantic-folder build [source_scope]`
- Agent loop: `pifs ask "<question>"`, `pifs chat`

The agent command surface intentionally exposes only read/navigation commands: `ls`, `tree`, `find`, `browse`, `grep`, `cat`, and `stat`. It can use an existing semantic folder like any other tree, but it cannot build one.

## Key Files

- `pageindex/filesystem/core.py`: high-level PIFS API, registration flow, metadata generation, projection wiring, semantic-folder build orchestration, and browse behavior.
- `pageindex/filesystem/store.py`: SQLite workspace catalog for folders, files, metadata, generated memberships, and PageIndex/projection state.
- `pageindex/filesystem/commands.py`: command parser, executor, shell rendering, capabilities, and guardrail messages.
- `pageindex/filesystem/agent.py`: `ask` / `chat` policy and streaming loop over the PIFS command surface.
- `pageindex/filesystem/semantic_folder.py`: Semantic Folder planner contract, OpenAI planner, plan schema, and validation rules.
- `pageindex/filesystem/semantic_projection.py` and `semantic_index.py`: summary projection indexing and vector search adapter used by `browse`.
- `pageindex/filesystem/metadata.py` and `metadata_generation.py`: metadata schema, policy, status, and generated metadata helpers.
- `pageindex/filesystem/cli.py` and `pifs`: CLI entrypoints.
- `examples/pifs_demo.py`: local end-to-end demo over example documents.

## Verification

- [x] `uv run pytest tests/test_filesystem_store.py tests/test_import_surface.py tests/test_metadata_generation.py tests/test_pageindex_filesystem_scope.py tests/test_pageindex_structural_read.py tests/test_pifs_add_command.py tests/test_pifs_agent_stream.py tests/test_pifs_cli.py tests/test_pifs_find_maxdepth.py tests/test_pifs_like_escape.py tests/test_pifs_path_resolution.py tests/test_pifs_register_side_effects.py tests/test_pifs_semantic_folder.py tests/test_semantic_index.py`
- [x] Manual PIFS CLI/chat demo coverage on the example workspace during development.
- [x] Manual Semantic Folder smoke on SEC filings: `/SEC_Filings_LTM/semantic` built as `topic/domain` with 82 files, 82 memberships, and 0 skipped files.
- [x] Manual Semantic Folder smoke on 33capital: `/33capital/semantic` built as `topic/domain` with 18 files, 18 memberships, and 0 skipped files.


pageindex/filesystem/commands.py
```diff
@@ -0,0 +1,1651 @@
...
+        relation = None
+        limit = 10
+        file_type = None
+        max_depth = None
+        i = 0
+        while i < len(args):
+            arg = args[i]
+            if arg == "--where":
+                i += 1
+                where = args[i]
```
KylinMountain: **[High] Malformed command crashes the agent turn / CLI instead of returning a recoverable error.** A value-consuming option with no value — `find /docs --where`, `--name`, `-type`, `--limit`; `grep ... --where`; `tree --depth` — runs `i += 1; x = args[i]` here with no bounds check and raises `IndexError`. `execute()` (L106-109) only catches `PIFSCommandError`/`KeyError`/`ValueError`, and `agent._run_bash` (L636-640) only catches `PIFSCommandError`, so the `IndexError` escapes: the CLI exits non-zero and the agent turn aborts — exactly the failure mode the "errors come back as ERROR: text" design is meant to prevent.

_Fix:_ bounds-check before `args[i]` (as `-maxdepth` already does), or also catch `IndexError` in `execute()`/`_run_bash`.

pageindex/filesystem/cli.py
```diff
@@ -0,0 +1,403 @@
...
+    _load_env_file(args.env_file, workspace=args.workspace)
+    args.workspace = _resolve_workspace(args.workspace)
+    if not args.workspace:
+        parser.error("--workspace is required unless PIFS_WORKSPACE is set or `pifs set workspace <path>` has been run")
+    return args
+
+
+def _filesystem_from_workspace(workspace: str) -> PageIndexFileSystem:
+    filesystem = PageIndexFileSystem(Path(workspace).expanduser())
+    filesystem.configure_existing_projection_retrieval()
```
KylinMountain: **[High] Read-only commands require an embedding API key once the workspace has indexed content.** `_filesystem_from_workspace` calls `configure_existing_projection_retrieval()` unconditionally; when a non-empty summary projection exists this eagerly builds `EmbeddingClient`, which raises `ValueError("PIFS_EMBEDDING_API_KEY or OPENAI_API_KEY is required ...")` in `__init__`. So `pifs ls/tree/find/grep/cat/stat` abort with no key set, even though they never embed. (An empty index returns early, so it only breaks after content is indexed.)

_Fix:_ construct the embedder lazily — only on `browse` / semantic retrieval.

pageindex/filesystem/store.py
```diff
@@ -0,0 +1,2416 @@
...
+    def search_files(
+        self,
+        query: str | list[str] | None,
+        *,
+        scope: Optional[dict[str, Any]] = None,
+        metadata_filter: Optional[dict[str, Any]] = None,
+        limit: int = 10,
+    ) -> list[dict[str, Any]]:
+        query_text = self._query_text(query)
+        match_queries = self._fts_match_queries(query_text) if query_text else [None]
```
KylinMountain: **[High] A stopword/punctuation-only query silently returns `[]`.** For a truthy query like `the`, `_fts_match_queries` strips all stopwords → `[]`, so `match_queries` is empty and the result loop never runs (returns `[]`). `query=None` instead uses the `[None]` fallback and returns the scope-filtered set. Net: `find /docs --name "the"` returns nothing even though scope + metadata filters alone would match.

_Fix:_ fall back to `[None]` when a non-empty query yields zero FTS terms.

pageindex/filesystem/core.py
```diff
@@ -0,0 +1,2439 @@
...
+        title = str(
+            self.store.membership_display_name(file_ref, folder_path)
+            or getattr(entry, "title", "")
+            or ""
+        ).strip()
+        if not title:
+            raise RuntimeError(f"browse cannot build a virtual path for {file_ref}: missing title")
+        target = self._join_virtual_file_path(folder_path, title.strip("/"))
+        try:
+            resolved_file_ref = self.store.resolve_file_ref(target)
```
KylinMountain: **[High] One ambiguous title aborts the entire `browse`.** `_stable_file_locator` calls `resolve_file_ref` and re-raises its `KeyError("Ambiguous file target")` as `RuntimeError`; in `browse_semantic_files` this call (~L973) is not wrapped in try/except, so a single candidate whose reconstructed virtual path resolves to >1 file_ref kills the whole query. Reachable when a title is duplicated across folders in scope (e.g. a source folder + its semantic-folder membership with a per-folder `display_name`).

_Fix:_ catch and `continue` past ambiguous candidates.

pageindex/filesystem/store.py
```diff
@@ -0,0 +1,2416 @@
...
+        operator, expected = next(iter(condition.items()))
+        field_id = self.field_id(field)
+        if operator == "$eq":
+            return (
+                """
+                EXISTS (
+                    SELECT 1 FROM metadata_values mv
+                    WHERE mv.file_ref = f.file_ref
+                      AND mv.field_id = ?
+                      AND mv.value_text = ?
```
KylinMountain, line 1059: **[High] Numeric `$eq/$ne/$in` compare `str()` forms, so int/float encodings never match.** A value stored as float `2024.0` (value_text `'2024.0'`) is excluded by `{"year": {"$eq": 2024}}` (compares `'2024'`), and `{"$eq": 2024.0}` misses an int-stored `2024`; `$ne` wrongly includes them. One of three holes in metadata comparison (see also the `$gt/$lt` notes).

_Fix:_ compare numerically when either side is numeric (use `value_number`).

KylinMountain (commented):
## Automated deep-review summary (recall-oriented)

Machine-assisted review (multi-agent finder + adversarial verification passes) of the PIFS change, optimized for **recall** of real defects. Each inline finding was independently verified against the PR-head code. Severity tags: **High** = breaks normal usage / crashes / silent wrong results; **Med** = real bug with a narrower trigger.

### Positioning question first (see the requirements.txt comment)
PIFS introduces `sqlite-vec` + an OpenAI embedding pipeline, and `pifs browse` performs **vector similarity search**, which sits in tension with the README's **"Vectorless / No Vector DB"** positioning (L14 / L16 / L65 / L67). Flagging for an explicit product + docs decision, not as a code bug.

### Highest-impact correctness issues (inlined below)
- Malformed command (`find /docs --where`) → uncaught `IndexError` crashes the agent turn / CLI.
- Read-only commands (`ls`/`cat`/`find`...) require an embedding key once content is indexed.
- Stopword-only query (`--name "the"`) silently returns `[]`.
- One ambiguous title aborts the whole `browse`.
- Metadata filters: numeric `$eq` int/float mismatch, `$gt/$lt` excludes text-stored numerics, large-int float precision loss — three distinct holes in the same area.

### Investigated but NOT flagged (verified safe)
`SQLiteSession` thread / shared-history concerns (SDK uses `check_same_thread=False` + per-instance in-memory DB); `decode_vector` dimension (index-level `search`/`upsert` already validate); `__init__.py` only swallows `ModuleNotFoundError` for the 4 optional deps with a re-`raise` guard; `add_file` readiness guards not reused on `register` (intentional deferred-metadata design); the folder-vs-file depth `-1` asymmetry (actually correct — a file is one tree level below its folder).

### Lower-severity (not inlined, for completeness)
`cat --range` past EOF → nonsensical `start>end` empty result; the pagination "next" command can point past document EOF; `SQLiteVecSemanticIndex.reset()` uses auto-committing `executescript` with no rollback (recoverable — rebuildable index); `with self.connect()` commits but never closes the connection (bounded by CPython GC); `entity`/`relation` channels are declared & plumbed but never populated (dormant — hidden from capabilities & rejected up front, only over-advertised in static help). The default-model change to `gpt-5.4` looks intentional (`retrieve_model` was already `gpt-5.4`).

There are also several **cleanup/DRY** opportunities (OpenAI-client construction duplicated 3x; `normalize_path` / JSON-coercion / `textwrap.shorten` reimplemented; `MetadataGenerator` lazy-init copy-pasted 3x; per-file LLM/embedding calls that could batch) — happy to file separately if useful.

_Generated with assistance; treat as input, not gospel — verify before acting._

pageindex/filesystem/commands.py
```diff
@@ -0,0 +1,1651 @@
...
+            elif arg.startswith("-"):
+                raise PIFSCommandError(f"Unsupported tree option: {arg}")
+            else:
+                path = arg
+            i += 1
+        if depth < 1:
+            raise PIFSCommandError("tree --depth must be at least 1")
+        if depth > self.MAX_TREE_DEPTH:
+            depth = self.MAX_TREE_DEPTH
+        listing = self.filesystem.browse(path, recursive=True, limit=limit)
```
KylinMountain: **[Med-High] `tree --depth` is ignored in `--json`, and `--limit` is applied before the depth cut.** `_cmd_tree` never passes `max_depth` to `browse()` (which accepts it); depth is only applied in the shell renderer (`_render_tree`). So `tree /docs --depth 1 --json` returns the whole subtree, and `tree --depth 1 --limit N` can spend the row LIMIT (`ORDER BY path`) on deep folders and drop depth-1 children from both JSON and the rendered tree.

_Fix:_ pass `max_depth` into `browse`.

pageindex/filesystem/commands.py
```diff
@@ -0,0 +1,1651 @@
...
+                    max_depth=max_depth,
+                )
+            folders = self.filesystem.browse(
+                path,
+                recursive=True,
+                limit=limit,
+                max_depth=max_depth,
+            )["folders"]
+            if max_depth is not None and limit != 0:
+                return [self.filesystem.folder_info(path), *folders][:limit]
```
KylinMountain: **[Med-High] `find -type d --where` returns a different folder set than `find -type d`.** The `--where` branch returns `find_folders(...)` (which excludes the start dir via `path != ?` and folders whose subtree has 0 matches), while the non-where branch prepends `folder_info(path)` and lists folders regardless. So `find /documents -maxdepth 1 -type d --where {...}` drops `/documents` itself and zero-match subfolders — inconsistent with the same flag without `--where`.

pageindex/filesystem/store.py
```diff
@@ -0,0 +1,2416 @@
...
+            SELECT
+                f.file_ref,
+                f.external_id,
+                f.title,
+                f.descriptor,
+                f.pageindex_tree_status,
+                f.metadata_json,
+                f.metadata_status_json,
+                f.created_at,
+                MIN(pf.folder_id) AS folder_id,
```
KylinMountain: **[Med] Multi-folder files get a mismatched (folder_path, folderId, title) tuple in listings.** `_file_rows_for_scope` selects `MIN(folder_id)`, `MIN(folder_path)`, `MIN(display_title)` as *independent* aggregates under `GROUP BY file_ref`, so for a file in >1 folder each field can come from a different membership. This is the normal case after `semantic-folder build` (file lives in its source folder + the semantic folder, with a per-folder `display_name`), so `ls`/`browse` can show folder_path from one folder and name/folderId from another.

_Fix:_ pick one membership row via a correlated subquery / window function, as `_search_once` does.

pageindex/filesystem/store.py
```diff
@@ -0,0 +1,2416 @@
...
+            "raw_artifact_path",
+            "pageindex_doc_id",
+            "pageindex_tree_status",
+            "metadata_json",
+            "metadata_status_json",
+        ]
+        columns.extend(["deleted_at", "updated_at"])
+        placeholders = ", ".join(["?"] * (len(columns) - 2) + ["NULL", "CURRENT_TIMESTAMP"])
+        return f"""
+            INSERT OR REPLACE INTO files ({", ".join(columns)})
```
KylinMountain: **[Med] `INSERT OR REPLACE` resets `created_at` and resurrects soft-deleted files on re-register.** The column list omits `created_at` and sets `deleted_at = NULL`; re-registering an existing `file_ref` (same external_id, or same folder+title → same deterministic ref) row-replaces, so `created_at` reverts to `CURRENT_TIMESTAMP` (breaks recency/audit ordering) and a previously soft-deleted file reappears.

_Fix:_ `ON CONFLICT ... DO UPDATE` preserving `created_at`/`deleted_at`, or SELECT-then-merge.

pageindex/filesystem/core.py
```diff
@@ -0,0 +1,2439 @@
...
+        records = [self._prepare_file_record(file) for file in files]
+        try:
+            for record in records:
+                self._generate_register_metadata(record)
+            self._register_generation_policy_schema(records)
+            self.store.insert_files(records)
+        except Exception:
+            self._cleanup_failed_register_artifacts(records)
+            raise
+        for record in records:
```
KylinMountain: **[Med] `register_files` can leave the catalog and vector index inconsistent on a partial failure.** The catalog commit (`insert_files`) is inside a try/except that only cleans up file artifacts; the summary-projection upsert + `update_file_metadata_status` run *after* it with no compensating cleanup (unlike `add_file`, which rolls back both catalog and projection). If `update_file_metadata_status`/`_sync_owned_raw_artifact` raises (or the process dies) after a vector was committed, the catalog says "not ready" while the vector exists.

_Fix:_ mirror `add_file`'s cleanup over the projection phase.

pageindex/filesystem/store.py
```diff
@@ -0,0 +1,2416 @@
...
+            }[operator]
+            if isinstance(expected, (int, float)) and not isinstance(expected, bool):
+                return (
+                    f"""
+                    EXISTS (
+                        SELECT 1 FROM metadata_values mv
+                        WHERE mv.file_ref = f.file_ref
+                          AND mv.field_id = ?
+                          AND mv.value_number IS NOT NULL
+                          AND mv.value_number {comparator} ?
```
KylinMountain: **[Med] Numeric range filters exclude values stored as text.** `$gt/$gte/$lt/$lte` require `value_number IS NOT NULL`, but `value_number` is NULL whenever the value was written as a Python `str`. So `"3"` is dropped by `{"version": {"$gte": 2}}` even though `3 >= 2`. Reachable because generated metadata values are stored uncoerced.

_Fix:_ coerce numeric-looking text, or compare on a normalized numeric form.

pageindex/filesystem/store.py
```diff
@@ -0,0 +1,2416 @@
...
+            items = []
+            for item in value:
+                items.extend(SQLiteFileSystemStore._metadata_value_items(item))
+            return items
+        value_json = json.dumps(value, ensure_ascii=False, sort_keys=True)
+        value_text = SQLiteFileSystemStore._metadata_compare_text(value)
+        return [
+            {
+                "value_text": value_text,
+                "value_number": float(value) if isinstance(value, (int, float)) and not isinstance(value, bool) else None,
```
KylinMountain: **[Med] Large integers lose precision (float REAL coercion) in range filters.** `value_number = float(value)` and `$gt/$lt` bind `float(expected)`, so `9007199254740993` is stored as `...992.0`; `{"$gt": 9007199254740992}` wrongly excludes it and `{"$lte": ...992}` wrongly includes it — wrong results for epoch-ms / snowflake-ID metadata.

_Fix:_ keep integers as INTEGER, or compare with awareness of the int/float distinction.

pageindex/filesystem/semantic_projection.py
```diff
@@ -0,0 +1,624 @@
...
+
+def embedding_cache_model_key(model: str, dimensions: int) -> str:
+    return f"{model}:dimensions={dimensions}" if dimensions > 0 else model
+
+
+def embed_with_retry(embedder: Any, texts: list[str], *, max_attempts: int = 8) -> list[list[float]]:
+    for attempt in range(1, max_attempts + 1):
+        try:
+            return embedder.embed(texts)
+        except Exception:
```
KylinMountain: **[Med] `embed_with_retry` retries *all* exceptions 8x with backoff, turning permanent errors into multi-minute hangs.** The broad `except Exception` + exponential sleep (capped 120s/attempt) retries a 401 (bad/missing key) or a code-level `TypeError` ~8 times (~127s cumulative) before re-raising. A bad key during register/add/browse hangs for minutes instead of failing fast.

_Fix:_ don't retry non-transient errors (auth / 4xx / `TypeError`).

pageindex/filesystem/store.py
```diff
@@ -0,0 +1,2416 @@
...
+        digest = hashlib.sha1(name.encode("utf-8")).hexdigest()[:16]
+        return f"field_{digest}"
+
+
+def normalize_path(path: str | Path | None) -> str:
+    if path is None:
+        return "/"
+    if str(path).strip().lower() == "root":
+        return "/"
+    parts = [part for part in str(path).replace("\\", "/").split("/") if part and part != "."]
```
KylinMountain, line 1969: **[Med] `normalize_path` keeps `..` segments, unlike sibling validators.** It drops `''` and `'.'` but not `'..'`; `_validate_add_filename` and `semantic_folder._validate_segment` reject `'..'`, but the `register_file` path runs only `normalize_path`. So `register_file(folder_path='/a/../b')` creates a literal `/a/../b` folder — phantom folders that break descendant-LIKE / depth math (which treat `..` as an ordinary segment).

_Fix:_ resolve or reject `..` in `normalize_path`.

pageindex/filesystem/semantic_projection.py
```diff
@@ -0,0 +1,624 @@
...
+                            model,
+                            hashes[index],
+                            len(vector),
+                            encode_vector(vector),
+                        )
+                        for index, vector in zip(positions, vectors)
+                    ],
+                )
+                conn.commit()
+            for index, vector in zip(positions, vectors):
```
KylinMountain, line 317: **[Med] `EmbeddingCache.embed_texts` zips positions↔vectors with no length check.** If `embed_with_retry` returns a different count than the texts sent (a custom/compatible embedder via `PIFS_EMBEDDING_*`, or a partial response), `zip` silently misaligns — vectors get persisted under the wrong `text_hash` and served for the wrong text on later hits, and `[cached[h] for h in hashes]` may `KeyError`.

_Fix:_ assert `len(vectors) == len(batch_texts)` before caching.

requirements.txt
```diff
@@ -1,6 +1,7 @@
 litellm==1.83.7
-# openai-agents  # optional: required for examples/agentic_vectorless_rag_demo.py
+# openai-agents==0.17.2  # optional: required for pifs chat/ask and examples/agentic_vectorless_rag_demo.py
 pymupdf==1.26.4
 PyPDF2==3.0.1
 python-dotenv==1.2.2
 pyyaml==6.0.2
+sqlite-vec>=0.1.9
```
KylinMountain, line 7: **[Positioning / architecture — for discussion] New hard dependency `sqlite-vec` + embedding-based similarity search.** PIFS adds `sqlite-vec` and an OpenAI embedding pipeline (`semantic_projection.py` / `semantic_index.py`), and `pifs browse` performs vector similarity search. That's in tension with PageIndex's headline positioning — the README states **"Vectorless, Reasoning-based RAG"** and **"No Vector DB ... instead of vector similarity search"** (README L14 / L16 / L65 / L67).

Worth an explicit product call: is the semantic projection an intentional, clearly-scoped file-*location* aid (with reasoning-based structural retrieval still primary), and should that be documented so it doesn't contradict the "vectorless" claim? It also makes `sqlite-vec` a required dependency.

pageindex/filesystem/commands.py
BukeLy: Fixed in f4fde16: malformed value-consuming options now use a shared bounds-checked parser and return PIFSCommandError instead of leaking IndexError. Resolving.

pageindex/filesystem/store.py
BukeLy: Fixed in f4fde16: stopword-only name searches now fall back to literal LIKE matching instead of returning an empty FTS query result. Resolving.

pageindex/filesystem/commands.py
BukeLy: Fixed in f4fde16: tree now passes max_depth into browse so JSON and rendered output honor --depth before limit. Resolving.

pageindex/filesystem/cli.py
BukeLy: Fixed in f4fde16: workspace open no longer eagerly constructs projection retrieval; existing projection retrieval is configured lazily when browse needs it, so read-only commands do not require an embedding key. Resolving.

pageindex/filesystem/core.py
BukeLy: Fixed in f4fde16: browse now isolates per-candidate virtual-path failures and skips an ambiguous candidate instead of aborting the whole result set. Resolving.

pageindex/filesystem/store.py
BukeLy: Fixed in f4fde16: numeric equality, inequality, and in-list filters now use decimal comparison so int, float, and numeric text encodings compare consistently. Resolving.

pageindex/filesystem/commands.py
BukeLy: Fixed in f4fde16: find -type d --where now uses the same bounded directory semantics as find -type d, including the start folder when maxdepth is bounded. Resolving.

pageindex/filesystem/store.py
BukeLy: Fixed in f4fde16: file writes now use ON CONFLICT DO UPDATE, preserving created_at and existing deleted_at tombstones instead of replacing the row. Resolving.

pageindex/filesystem/store.py
BukeLy: Fixed in f4fde16: numeric range filters now compare decimal value_text, so numeric strings such as "3" participate in numeric ranges. Resolving.

pageindex/filesystem/store.py
BukeLy: Fixed in f4fde16: range filters no longer coerce expected values through REAL; Decimal comparison preserves large integer precision. Resolving.

pageindex/filesystem/core.py
BukeLy: Fixed in f4fde16: register_files now compensates new catalog rows and summary projection entries if a post-catalog step fails, keeping catalog and projection state consistent. Resolving.

pageindex/filesystem/store.py
BukeLy: Fixed in f4fde16: recursive listings now select one consistent file_folders membership row, so folder_id, folder_path, and display title cannot come from different rows. Resolving.

pageindex/filesystem/semantic_projection.py
BukeLy: Fixed in f4fde16: embedding retry now only backs off for retryable status/classes and immediately surfaces permanent 4xx errors such as 401. Resolving.

pageindex/filesystem/semantic_projection.py
BukeLy: Fixed in f4fde16: EmbeddingCache now validates response vector count before zipping positions to vectors, preventing wrong hash-to-vector cache entries. Resolving.

pageindex/filesystem/store.py
BukeLy: Fixed in f4fde16: PIFS path normalization now rejects parent-directory segments so register_file cannot create phantom folders through /a/../b paths. Resolving.

requirements.txt
BukeLy: Not changing this in the correctness-fix commit. The code fixes above keep semantic projection scoped to PIFS `browse` as a ranked file-location aid; the semantic backend returns candidate document ids for catalog resolution, while evidence still comes from bounded `cat`/`grep`/`stat`/PageIndex reads. The README/product wording needs an explicit docs decision rather than a silent dependency edit in this bugfix pass, so I am leaving this as positioning context for the PR discussion.

BukeLy:
Thanks for the contribution. Closing this PR because the relevant indexing behavior has been superseded by the current Flash pipeline, PageIndex File System direction, or the later documentation/assets implementation. The branch is no longer suitable to merge as-is.

PR #211, opened 2026-04-02
Add packaging config: dependency bounds, console script, and pyproject-based install

trehansalil:
This pull request sets up the initial project configuration for a new Python package named `pageindex`. The main changes include specifying Python version requirements and adding project metadata and dependencies.

Project setup and configuration:

* Added a `.python-version` file to specify Python 3.12 as the required version for development.
* Created a `pyproject.toml` file with project metadata, Python version requirement (>=3.12), dependencies (`litellm`, `pymupdf`, `PyPDF2`, `python-dotenv`, `pyyaml`), and build system configuration using Hatchling.

PR #348, opened 2026-07-08
ci: publish to PyPI on version tags

KylinMountain:
Adds a tag-triggered PyPI release workflow directly on `main`, so the stable branch owns the release infrastructure independently of `dev`.

## What it does
Pushing a PEP 440 tag with a leading `v` (`v0.3.0`, `v0.3.0rc1`, `v0.3.0.dev2`, …) triggers `.github/workflows/publish.yml`, which:
1. derives the version from the tag and injects it into `pyproject.toml`;
2. builds sdist + wheel;
3. publishes to PyPI via **OIDC trusted publishing** (no stored token/secret);
4. creates a GitHub Release with auto-generated notes.

## Scope
Single file: `.github/workflows/publish.yml`. No code changes.

Supersedes the dev-based #347 (same workflow, targeted at `main` instead).

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,63 @@
+name: Publish to PyPI
+
+# Release flow (the git tag IS the version — nothing to bump in the repo):
+#   1. git tag -a v0.3.0.dev2 -m "Release 0.3.0.dev2"
+#   2. git push origin v0.3.0.dev2
+#   3. This workflow derives the version from the tag, injects it into
+#      pyproject.toml, builds, publishes to PyPI via OIDC trusted publishing
+#      (no stored secret), and creates a GitHub Release with generated notes.
+#
+# The tag must be a PEP 440 version with a leading `v`:
+#   v0.3.0   v0.3.0rc1   v0.3.0.dev2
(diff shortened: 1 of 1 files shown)


rejojer:
### Code review

Found 1 issue:

1. **Workflow will always fail: `pyproject.toml` does not exist on `main`** (bug: missing build dependency on target branch)

The workflow's build step assumes `pyproject.toml` exists at the repo root, but `main` has no `pyproject.toml`, `setup.py`, or `setup.cfg` — only a flat `requirements.txt`. The file only exists on `dev`. Under `set -euo pipefail`, `sed -i` will fail immediately with "No such file or directory" on every tag push, making the workflow completely non-functional.

This appears to have been copied from PR #347 (which correctly targets `dev`) without accounting for the packaging divergence between `main` and `dev`. To fix, either: (a) add `pyproject.toml` with a `[build-system]` table to `main` before or as part of this PR, (b) retarget this PR to `dev`, or (c) merge one of the pending packaging PRs (e.g. #211) to `main` first.

https://github.com/VectifyAI/PageIndex/blob/e048991914346ee9d196fabe0fffcb5d821dfe51/.github/workflows/publish.yml#L43-L52

KylinMountain:
closed as #347 has merged into dev.

PR #356, opened 2026-07-16
Add Atlas Cloud LiteLLM provider alias

binyangzhu000-sudo:
## Summary
- add an `atlascloud/` model alias that routes PageIndex LiteLLM calls through Atlas Cloud's OpenAI-compatible endpoint
- read `ATLASCLOUD_API_KEY` and optional `ATLASCLOUD_API_BASE` only when Atlas Cloud models are selected
- document a minimal Atlas Cloud config example and add focused unit coverage for the model mapping

## Validation
- `.venv/bin/python -m compileall pageindex`
- `.venv/bin/python -m pytest tests/test_atlascloud_litellm.py tests/test_issue_163.py`
- `git diff --check`
- verified Atlas Cloud public model catalog contains `qwen/qwen3.5-flash` and `deepseek-ai/deepseek-v4-pro`

No sponsor, logo, credits, or partner sections were added.


binyangzhu000-sudo:
Resolved the merge conflicts with the current `main` in `9d6af7c`. The resolution preserves the new direct OpenAI SDK routing while keeping `atlascloud/...` models on LiteLLM with the Atlas Cloud endpoint and API key.

Validation:
- `python 3.11 -m pytest -q tests` (`25 passed`)
- `git diff upstream/main --check`

binyangzhu000-sudo:
Synced with `main` (56 commits behind) and rebuilt this on the reworked LLM lane rather than restoring the old shape.

**What upstream changed, and how this adapts**

- The direct OpenAI-SDK route (`_is_openai_model`, `use_openai_sdk`) is gone and everything goes through LiteLLM now. I took that deletion as-is — the Atlas support no longer rides on that path. (Checked it was yours to delete, not mine to keep: `_is_openai_model` was already in this PR's merge base, while `prepare_litellm_call` was the part I'd added.)
- `prepare_litellm_call()` returned `(model, kwargs)`, which no longer matches anything. Split it to fit your shape instead: `_litellm_model()` gets an `atlascloud/` branch that normalizes to the `openai/` wire form, and a new `_atlascloud_kwargs()` supplies api_base/api_key. `llm_completion` and `llm_acompletion` merge those **before** `_llm_backend`, so an explicit backend override still wins.
- Two test assertions updated for upstream behaviour changes, not for my code: bare model names now get the `openai/` prefix, and completion kwargs carry `max_retries: 0` instead of `temperature: 0`.

**README**: taken wholesale from upstream. My old snippet used the retired `model:`/`retrieve_model:` YAML keys and the page is now SDK-oriented with no provider-configuration section — dropping it in anyway would have been noise. The runnable example stays in `pageindex/config.yaml`, which auto-merged onto your current `index_model`/`chat_model` naming.

**Verification** — `pytest tests/`: **330 passed, 162 skipped**, including this PR's 8 Atlas cases.

Disclosure: I work at Atlas Cloud.

PR #357, opened 2026-07-17
Improve JSON fence parsing

adity982:
## Summary
- support case-insensitive JSON code fences when extracting LLM JSON responses
- support plain Markdown code fences in get_json_content
- add regression tests for uppercase and plain fences

## Validation
- python -m pytest tests/test_page_index.py


BukeLy:

PR #378, opened 2026-08-02
Trim the Flash README

rejojer:
Trims the Flash package README and corrects what it claims.

The opening said "No LLM, no API key, no OCR, no network", directly above a
usage example that calls an LLM: `summary=True` is the default, so
`page_index_flash("paper.pdf")` needs a key. Without one it does not raise
either, it retries and returns empty summaries. The first line now says which
part needs a model and which does not, and the usage block shows all three
calls.

Also in this pass:

- Usage split into Python and command line, each with a line saying what you
  get back and where it goes.
- The output schema gains `has_abstract_or_references_section` and `summary`,
  both of which the function actually returns, and drops the `--optimize` CLI
  spelling from what is a Python dict.
- The cost figure, the per-1,000-page and per-node dollar figures and the token
  pricing line are gone, along with `assets/cost_vs_pages.png`. Token counts
  stay; they do not go stale when a price does.
- Limits and Dependencies dropped.

74 lines to 71, and nothing in it is now contradicted by the code.


rejojer:
Superseded by #379, opened from a fresh branch so the diff is not buried under the old release branch's history.

PR #389, opened 2026-08-06
feat: local mode for the PageIndex SDK (v0.2.9)

rejojer:
## What

Local mode for the PageIndex SDK. `PageIndexClient` keeps the exact 0.2.8 cloud surface and gains a fully local backend (standard + Flash indexing) — no server, no API key, results stored as JSON on disk.

```python
from pageindex import PageIndexClient, PageIndexCloudClient, PageIndexLocalClient

client = PageIndexClient(api_key="...")   # cloud (0.2.8 behavior)
client = PageIndexClient()                # local
# or pin the mode explicitly:
client = PageIndexCloudClient(api_key="...")
client = PageIndexLocalClient(storage_path="./.pageindex")
```

The 0.3.0.devN pre-releases explored a collection-based design; this release stays on the collection-free 0.2.x line, hence **0.2.9**.

## Cloud half

`pageindex/cloud_api.py` is the published 0.2.8 client kept line-for-line, plus only:

- request timeouts (uploads still unbounded)
- upload file handle closed via `with`
- URL-encoding of ids in request paths
- empty-body DELETE responses handled
- streaming: empty-`choices` guard, response closed on exit
- optional `submit_document(..., metadata={...})` (form field the server already accepts)
- `get_tree`: `include_text` query param now wired to the server (was accepted but never sent)
- `get_page_content`: raises `PageIndexAPIError` instead of `TypeError` when the document is still processing

## Local half

- `submit_document(pdf, mode="standard" | "flash")` runs PageIndex in-process; synchronous, returns a completed doc_id
- Storage: one directory per doc under `storage_path` (default `./.pageindex`) — `tree.json`, `pages.json`, `doc.json`; atomic writes; `doc.json` written last on save and removed first on delete, serving as the completeness marker
- `manifest.json`: write-through cache of all document metas for fast listing — self-heals from the per-doc files, per-entry validation, no locks (1000-doc listing ~8 ms warm)
- Models come from packaged defaults, overridable per client (`model`, `summary_model`, `retrieve_model`)

## Parity: aligned / different / cloud-only

**Aligned** — local mirrors the cloud wire shapes:

- doc ids: `pi-` + 32 hex chars
- `createdAt`: naive UTC, millisecond precision
- tree nodes: `page_index` (not `start_index`), non-leaf `prefix_summary`, `text` present
- envelopes: `{doc_id, status, retrieval_ready, result, metadata, features}`; list: `{documents, total, limit, offset}`; delete: `{"message": "Document deleted successfully."}`
[...]

PR #391, opened 2026-08-08
Flash embedded-TOC merge: dissolve duplicated headings, place new sections at their level

rejojer:
Two fixes to the FULL-tier bookmark merge in `pageindex/flash/embedded_toc.py`. Extracted from the sdk-local line of work so #389 stays SDK-only — flash is independent of the SDK changes.

## 1. Duplicated headings survived the merge (bf5d841)

The merge only same-heading-checked the single deepest bookmark anchor on a page. When a detected heading duplicated a *shallower* same-page bookmark — typically a numbering-prefixed variant (`5 Experiments` vs bookmark `Experiments`) — it grafted inside its own subsection, dragging its detected children with it.

Now a detected node checks the whole same-page frame chain for an identity match (normalized-equal / numbering-prefix suffix / fuzzy ≥ 0.9), dissolves into the match, and its children re-place individually. The fuzzy bar is 0.9, deliberately above the 0.7 title-repair bar: a false dissolve deletes a real section (0.7 killed the annual report's real "Reserve Bank and Branch Directors", 0.746 vs its chapter title), while OCR garbles that need fuzzy sit around 0.95 ("Applications of peA").

## 2. New detected sections always nested under the deepest anchor (c9b9220)

A detected heading with no bookmark counterpart was always attached to the deepest open bookmark section at its page — so a paper's unbookmarked `Conclusion` / `References` landed *inside* `7 Related Work`.

Since sections are intervals, a new node's parent must lie on the open-section chain at its page; placement is a depth choice on that chain. The node now climbs the chain while there is positive evidence the candidate is a peer, not a parent:

- **numbering**: node and candidate both numbered and the node is not a numbering-child of the candidate (`2.1` stays under `2.`, `4` climbs past it), or
- **style-fingerprint equality**: identical dominant style hash (face + weight + size). Raw font size is deliberately *not* evidence — margin-note headings can be larger-but-regular vs smaller-but-bold real subsections, and size-based climbing hoisted exactly that noise to chapter level.

A same-page climb additionally requires a vertical gap ≥ 0.15 page heights below the candidate heading — climbing on a near-zero gap asserts the section is empty, which is the margin-box signature, not a real successor section. Bookmark frame nodes inherit style/position from duplicates that dissolve into them (fix 1 feeds fix 2), and the internal `_style`/`_y` keys are stripped from the final structure.

## Verification

Before/after structure snapshots over the nine example PDFs with `summary=False` (LLM-free, deterministic):

- Fix 1: 40 duplicate nodes removed, zero other tree changes.
- Fix 2: 6/9 trees byte-identical; every delta reviewed — `attention` `Conclusion`/`References` become top-level sections (the target), PRML margin bios all stay contained at their real depth, PRML ch7 exercise items re-level to the chapter per numbering (the bookmarks have no ch7 `Exercises` node; this corrects two sections' end pages).

8 unit tests added (`tests/test_flash_embedded_toc.py`); full suite 26 passing.

### DIFF
--- a/pageindex/flash/embedded_toc.py
+++ b/pageindex/flash/embedded_toc.py
@@ -51,6 +51,7 @@ IGNORE, SKELETON, FULL = 1, 2, 3
 
 # Same-page titles at or above this similarity refer to the same section.
 _REPAIR_SIMILARITY = 0.7
+_DISSOLVE_SIMILARITY = 0.9
 
 # Backfill noise filters: a normalized title recurring this often across the
 # detected tree is a running label, not a section; a title longer than this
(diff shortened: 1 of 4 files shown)


BukeLy:
Thanks for the contribution. Closing this PR because the relevant indexing behavior has been superseded by the current Flash pipeline, PageIndex File System direction, or the later documentation/assets implementation. The branch is no longer suitable to merge as-is.

PR #407, opened 2026-08-14
docs: drop the demo's install step — openai-agents ships with the SDK now

rejojer:
Ports the pending commit from `feat/local-chat` to `main` (the #402…#406 pattern; parity verified before opening). Completeness follow-up to #406: the README demo section still called openai-agents an optional dependency with a manual install step — it is a base dependency now, so the step is gone. After this merge, `git diff main feat/local-chat` is empty.


rejojer:
Not urgent — README rework deferred to the 0.2.10 stable release.

PR #442, opened 2026-08-28
docs: add an "Any MCP client" integration item

rejojer:
Adds one `<details>` item to **(b) Integrate PageIndex with your own agent**: connecting any MCP-speaking client to the hosted PageIndex MCP server, no SDK involved.

The config snippet matches https://docs.pageindex.ai/mcp verbatim (`https://api.pageindex.ai/mcp`, `type: http`, Bearer auth). The SDK-internal `?tools=read` gate is deliberately left out since the docs do not expose it. The item closes by pointing local users back to `agent_tools()` / `as_claude_mcp()`, so it never implies local mode has a standalone server.

Placed after "Other agent frameworks" so the "Every helper above accepts `doc_id=`…" wrap-up still covers only the SDK helpers. README-only, additive.

https://claude.ai/code/session_01F84VYrC858n9cr5eRjV4Cy

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,7 +285,7 @@ Returns the agent's process transcript in `items`. Append those items to the nex
 client.messages("...", model="claude-sonnet-4-6", doc_id=doc_id)
 ```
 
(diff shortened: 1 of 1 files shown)


rejojer:
Folded into #443 so the code and its README land together.

Issue #153, opened 2026-03-07
Unsupported operand type(s) for +: 'int' and 'NoneType'


anujbansal:
(venv) anujbansal@MacBook-Pro-2 PageIndex % python3 run_pageindex.py --pdf_path /Users/anujbansal/Dropbox/Mac\ \(2\)/Desktop/AI\ Research/A.\ C.\ Muller\ and\ S.\ Guido\ -\ Introduction\ to\ Machine\ Learning\ with\ Python\ -\ 2017.pdf
Parsing PDF...
start find_toc_pages
toc found
start detect_page_index
index found
process_toc_with_page_numbers
start_index: 1
start toc_transformer
start toc_index_extractor
Traceback (most recent call last):
  File "/Users/anujbansal/Dropbox/Mac (2)/Documents/workspace/PageIndex/run_pageindex.py", line 67, in <module>
    toc_with_page_number = page_index_main(args.pdf_path, opt)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/anujbansal/Dropbox/Mac (2)/Documents/workspace/PageIndex/pageindex/page_index.py", line 1100, in page_index_main
    return asyncio.run(page_index_builder())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.11/3.11.15/Frameworks/Python.framework/Versions/3.11/lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.11/3.11.15/Frameworks/Python.framework/Versions/3.11/lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.11/3.11.15/Frameworks/Python.framework/Versions/3.11/lib/python3.11/asyncio/base_events.py", line 654, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/Users/anujbansal/Dropbox/Mac (2)/Documents/workspace/PageIndex/pageindex/page_index.py", line 1075, in page_index_builder
    structure = await tree_parser(page_list, opt, doc=doc, logger=logger)
[...]

PR #466, opened 2026-09-02
fix: return 0 instead of None when calculate_page_offset has no valid pairs

LCARyanKing:
Fixes a crash when processing PDFs that have no matching page pairs between the table of contents page numbers and physical indices.

When calculate_page_offset() cannot determine an offset (no valid pairs), it now returns 0 instead of None, allowing add_page_offset_to_toc_json() to safely perform arithmetic operations.


LCARyanKing:
This is related to issue: https://github.com/VectifyAI/PageIndex/issues/153

PR #492, opened 2026-09-08
Tool-path rate limits: retry at the bridge, then fail the run fast

rejojer:
A PageIndex cloud 429 (or 5xx) on a tool call used to reach the model as an `INTERNAL_ERROR` envelope saying "try again": the model re-called once with no wait, then wrote the failure into its answer, and `chat()` returned normally with no status anywhere. The same 429 before the loop (the doc_id targeting lookup) already propagated raw.

## Changes

- **`McpBridge`** mounts a urllib3 `Retry`: 429, every 5xx and connection failures, three attempts, 0/2/4 s apart. `Retry-After` is ignored: a long one is a quota, not a blip, and parsing it was the one way a 429 could turn into "could not reach the server" (urllib3 raises on a non-integer value). Read timeouts are never replayed (240 s each, and the server may have acted). Exhausted, the last response falls through to the existing `>= 400` branch, so `status_code` survives.
- **`_bridge_invoker`** re-raises 429/5xx and an unreachable server (a transport failure that outlived the connection retries) alongside 401/403, and the cloud's two account-level tool errors, which arrive inside a normal HTTP 200 tool result with an `errorCode` (pageindex-chat #472): `RATE_LIMITED` (upstream 429, already retried server-side; `retry_after_seconds` when known) as 429 and `USAGE_LIMIT_REACHED` (plan quota, `open_url`) as 402. A non-JSON 200 body stays a model-visible envelope: the server was reached. The frameworks turn a raised tool exception back into model-visible text, so each `chat()` door gets its own escape:
  - openai-agents: the in-process `MCPServer` passes a `failure_error_function` that lets a PageIndex-caused failure propagate, and `_translate_run_error` unwraps it from the framework's wrapper (this also un-flattens the mid-session 401 case).
  - Messages lane: each turn's tools run through the runner's public `generate_tool_call_response()`, and a recorded failure raises before the next model call.
- **`_model_backend_error`** keeps the provider's `status_code`.
- The handshake error blames the API key only on 401/403; a rate-limited handshake no longer tells the user to rotate a working key.

The Claude Agent SDK lane and `as_openai_tools(hosted=True)` are out of reach: for a cloud client they hand the framework a URL and the framework's own MCP client makes the request, so neither the bridge retry nor `_bridge_invoker` is in the path.

## Behaviour change on the public tool surfaces

`as_openai_tools()` users running their own `Runner.run` now get an exception for a post-retry 429/5xx or an unreachable server (the 401/403 re-raise always intended this; the framework absorbed it). `as_anthropic_tools()` users' runners absorb it into an `is_error` result and log a traceback. The plain functions from `agent_tools()` raise `PageIndexAPIError` for the same set, where their docstring used to promise only 401/403; the docstrings now state the real raise set.

## Verification

- New tests against a local HTTP stub: the retry schedule (fixed backoff with a `Retry-After` header present and ignored), exhausted-retry status for 429, 504 and 529, read-timeout non-replay, an unreachable server re-raised through the invoker, the handshake wording per status; plus the invoker re-raise matrix (including the `RATE_LIMITED` / `USAGE_LIMIT_REACHED` envelopes and a non-JSON 200 body), openai-agents escape, run-error unwrap, provider `status_code`, end-to-end fail-fast on the chat and Messages doors, and a regression guard for model-side slips.
- Full suite green; the without-frameworks leg green; both agent test modules green on the openai-agents 0.18.1 floor; anthropic 0.108.0 already has `generate_tool_call_response`.
- Live against the cloud with real models: a forced post-retry 429 raises `PageIndexAPIError(status_code=429)` on `chat()` default, streamed, and `protocol="messages"`, after one tool call.

https://claude.ai/code/session_014S88dcSz7jykegAWyWZk8E
https://claude.ai/code/session_013xk3xt9KgHNTjsYmFLKxbu

https://claude.ai/code/session_01F7vqmZdnWeKC9SUBytDrdf

### DIFF
--- a/pageindex/agent_tools.py
+++ b/pageindex/agent_tools.py
@@ -1395,12 +1395,9 @@ def _bridge_invoker(bridge, name: str, schema: dict,
                          if value is not None}
             _coerce_bool_args(schema, arguments)
             blocks, is_error = bridge.call_tool(name, arguments)
-            if is_error:
-                _raise_account_limit(blocks)
-            return blocks, is_error
         except Exception as exc:
             if isinstance(exc, PageIndexAPIError) and (
-                    exc.status_code in (401, 402, 403, 429)
+                    exc.status_code in (401, 403, 429)
                     or (exc.status_code or 0) >= 500
                     or (exc.status_code is None and isinstance(
                         exc.__cause__, requests.RequestException))):
@@ -1414,6 +1411,9 @@ def _bridge_invoker(bridge, name: str, schema: dict,
                 "INTERNAL_ERROR",
             )
             return [{"type": "text", "text": _dumps(payload)}], True
+        if is_error:
+            _raise_account_limit(blocks)
+        return blocks, is_error
     return _invoke
 
 
--- a/pageindex/agent_tools.py
+++ b/pageindex/agent_tools.py
@@ -16,7 +16,8 @@ that envelope too, on the direct and the call_tool path alike, except
 browse_documents' ``recursive``: call_tool honors it, because the flat
 no-folders shape it asks for is trivially true here. The exceptions to
 never-raise, all cloud: a 401/403, a 429/5xx that outlived the bridge's
-retries and an unreachable server re-raise PageIndexAPIError.
+retries, an unreachable server and a RATE_LIMITED / USAGE_LIMIT_REACHED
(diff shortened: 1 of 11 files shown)


rejojer:
@codex review

PR #493, opened 2026-09-08
chat(protocol="chat_completions") and the compat note on chat_completions()

rejojer:
## Summary

- `chat()` gains a third protocol value, `"chat_completions"`: the answer lane's own engine with its Chat Completions envelope kept (chunk dicts when streaming, `instructions` as a leading system row). It is the one protocol the managed cloud chat serves, so that lane opens without a chat model; the own-model knobs still refuse there.
- `chat_completions()` stays public with its signature unchanged. Its docstring now marks it as kept for existing code and points new code at `chat()`, so everything it offers has a route that does not go through the old name. No warning, no `__getattr__` interception.
- The managed endpoint takes `extra_body` as its own request fields (`temperature`, `enable_citations`), merged into the payload last under the same skeleton refusal as the other lanes.
- Strings that steered callers to `chat_completions()` now name the protocol lane; the cookbook's two cells use `chat()`. Descriptive mentions are untouched.
- BREAKING: `chat()` takes keyword arguments after `messages`. Its positional order (`doc_id`, then `stream`) is the reverse of `chat_completions()`'s, and with the compat note inviting the rename, `chat(msgs, True, "pi-1")` bound `doc_id=True`: an unscoped, billed answer with no error. Now a `TypeError`. Nothing in the repo, the docs, or the cookbook passed `chat()` a positional after `messages`. The implementation and catch-all overload also carry the protocol `Literal`, so a misspelt protocol fails type-checking.

## Verification

- 511 passed; the new tests fail against main's code.
- Without-frameworks leg passes (`agents` blocked in `sys.modules`).
- pyright 236, same as main.

https://claude.ai/code/session_015b7YGZ8LvYp2Q3oGNN8Lfc
https://claude.ai/code/session_01DCBitejpuoNpD8cmQ9HR7j

### DIFF
--- a/cookbook/agentic_retrieval.ipynb
+++ b/cookbook/agentic_retrieval.ipynb
@@ -339,11 +339,7 @@
       "source": [
         "query = \"What are the evaluation methods used in this paper?\"\n",
         "\n",
-        "for chunk in pi_client.chat_completions(\n",
-        "    messages=[{\"role\": \"user\", \"content\": query}],\n",
-        "    doc_id=doc_id,\n",
-        "    stream=True\n",
-        "):\n",
+        "for chunk in pi_client.chat(query, doc_id=doc_id, stream=True):\n",
         "    print(chunk, end='', flush=True)"
       ]
     },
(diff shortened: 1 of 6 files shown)


rejojer:
@codex review

PR #494, opened 2026-09-09
MCP prompts over the bridge + chat(citations=True)

ahonn:
Two pieces, one stack: the MCP bridge learns to fetch server prompts, and `chat(citations=True)` is the first consumer, folded in from #496.

## Bridge: server prompts over the MCP session

Adds `prompts/list` and `prompts/get` to `McpBridge`, so the SDK can fetch the server's `cited_answer` prompt (with its optional `format` argument) over the existing MCP session.

- `list_prompts()` / `get_prompt(name, arguments=None)`, plus `render_prompt_text()` for system-prompt placement.
- Server capabilities are recorded at initialize; prompt calls fail with a clear error when `prompts` is not advertised.

Wiring into the `*_agent_config` bundles stays a follow-up.

## `chat(citations=True)`

Own-model chat cites every claim the way PageIndex chat does: `<cite doc="…" page="…"/>` tags, `block="…"` added where the cloud document has blocks.

- Own-model chat over cloud documents: fetches the server's `cited_answer` prompt (`format=cite`, the same rules PageIndex chat uses) over the same bridge session as the tools, and folds it into the system prompt after the managed prompt and before the caller's `instructions`. All three lanes (answer lane, `protocol="responses"`, `protocol="messages"`) via the existing instructions plumbing; `local_chat.py` only reworded one refusal.
- Managed chat: `citations=True` passes `enable_citations=True` to the endpoint. The endpoint cites in its own `<doc=…;page=…>` markup, not `<cite>` tags, and the resolved citations it returns (bbox where the document has blocks) come back only from `chat_completions`.
- Local documents: a frozen copy of the same prompt (`LOCAL_CITATION_PROMPTS`, the live text minus the one bullet naming the cloud-only `get_document_image()` tool). Local page content carries no blocks, so citations resolve to pages.
- `citations` takes only `True`/`False`. Another format goes through `client.citation_prompt(format=...)` passed as `instructions=`; the getter also serves framework-built agents. `markdown` / `footnote` are the prompt's variants for hosts that strip tags.

## Verified

- 503 tests green (3 new bridge tests, 9 new citation tests, red-verified first), without-frameworks leg simulated, pyright clean. Bridge smoke-tested against `api.pageindex.ai/mcp`.
- Key-gated live tests: the three formats served and distinct; local frozen copies equal the live text minus that bullet.
- Live with real models: block document → `<cite doc="document_5.pdf" page="1" block="p1_text_1"/>`; legacy document → `<cite doc="1706.03762_24.pdf" page="1"/>`; locally indexed PDF → `<cite doc="q1-fy25-earnings.pdf" page="3"/>`; managed `citations=True` non-stream and stream.

## Not in this PR

- Docs (`agents.mdx`, `chat.mdx`'s hand-written `<cite>` system message becomes `citations=`) and a release-notes line, with the release.
- Server-side, for the prompt's owner: the managed API's citation prompt carries a literal example id `p1_text_001` that the model echoes on legacy documents (resolved citations then carry a `block_id` with no `bbox`); and a `granularity` argument on `cited_answer` would let the SDK offer page-only citations on block documents. Two more: the managed prompt writes the `<doc=…;page=…>` form its parser labels Old while `cited_answer` writes `<cite …/>` (the parser takes both, so moving the managed prompt to `<cite>` gives one syntax across lanes); and `cited_answer`'s GROUNDING ("never fill a gap from general knowledge") contradicts the MCP instructions' general-knowledge exemption, which the managed prompt keeps.

https://claude.ai/code/session_01XGBdRF2qoTcLNpRahj6QHx

### DIFF
--- a/pageindex/agent_tools.py
+++ b/pageindex/agent_tools.py
@@ -34,7 +34,7 @@ from typing import Any, Callable, Optional
 import requests
 
 from .errors import PageIndexAPIError
-from .mcp_bridge import render_text
+from .mcp_bridge import render_prompt_text, render_text
 
 TOOL_RESPONSE_CHAR_LIMIT = 100_000
 STRUCTURE_FIRST_PAGE_THRESHOLD = 20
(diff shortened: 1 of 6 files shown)


rejojer:
@codex review

PR #495, opened 2026-09-09
Targeting moves to the first user message: document_context() replaces doc_id on the agent surfaces; folder_context() and chat(folder_id=) add folders

rejojer:
## What

The doc-targeting block ("The user has specified document: …") is conversation content, not guidance. It now lives in exactly one place on every surface: the first user message. Nothing is appended to the system prompt any more.

- **chat()**: all three lanes prepend the block as the first user message. The Messages lane moved off its separate system block; the OpenAI lanes already did this.
- **New `client.document_context(doc_id)`**: returns the block text for callers who own the conversation (the framework routes). Lead your first message with it:
  ```python
  agent = Agent(**client.openai_agent_config())
  Runner.run_sync(agent, [
      {"role": "user", "content": client.document_context(doc_id)},
      {"role": "user", "content": question},
  ])
  ```
  (Anthropic tool_runner and Claude Agent SDK: same text, prepended to the first message / `prompt`.)
- **BREAKING**: `doc_id` removed from eight methods — `agent_instructions()`, `openai_agent_config()`, `anthropic_runner_config()`, `claude_agent_config()`, `agent_tools()`, `as_openai_tools()`, `as_anthropic_tools()`, `as_claude_mcp()`. The prompt-side targeting moves to `document_context()`; the tool-layer allowlist was local-only and raised on cloud, so it is no longer a public knob (it stays internal to local `chat(doc_id=)`).
- **The block is O(1)**: one `get_document` per id. Names are unique per library (uploads suffix a taken name), so the same-name shadow check and its full-library listing sweep went (10 s → 1 s on a 749-doc cloud library). The cloud metadata endpoint carries `metadata`, so local `get_document()` gained that key for parity.
- **Rendering matches the cloud's managed chat**: same wording, one document as a JSON object, several as a list, the metadata row as returned.

## Folders (#499, rebased in)

The managed `/chat/completions` already takes `folder_id` and renders a folder block ahead of the document block (compute `api.py:491`, `:3182`); the SDK now sends it and mirrors it.

- **New `client.folder_context(folder_id)`**: the folder analogue of `document_context()`: the folder's name, an id/name/description metadata row (the server's `FolderContext`), and the directive to discover its documents with `browse_documents` / `search_documents(folder_id=..., recursive=true)`. One `list_folders()` call: the endpoint has no LIMIT, lists the same `userId` + `api_surface` partition the managed chat validates against, and carries the same plan gate, so a non-Max key fails here rather than mid-run. Cloud-only: a local client raises.
- **`chat(folder_id=)`** (also `chat_completions()` and the `responses` / `messages` lanes): the managed lane sends the field and lets the server render and gate it; the own-model lanes lead the conversation with one user message, folder block then document block, joined as the server joins them. `folder_id` joins the prompt-cache key like `doc_id`.
- `"root"` (and `""`) is the library itself: `folder_context` returns `""`, `chat` places nothing, on every lane and in every mode, as the server treats `folder_id="root"`.
- `folder_id` is keyword-only on `chat()` and trailing on every other signature: no positional call shifts.

## Why

System prompt = who you are and how you work. A per-request target there breaks the cached prefix, sticks across turns, and on cloud splices SDK text onto the server-served instructions. The cloud's own managed chat (compute `api.py`) and the web app both place document context in the user turn; this aligns the SDK with them.

## Verification

- `python -m pytest tests`: 480 passed. Without-frameworks CI leg simulated (agents / anthropic / claude_agent_sdk blocked): green.
- Live, real models, local store + cloud library × {chat() answer lane, `chat_completions()`, `protocol="responses"`, `protocol="messages"`, OpenAI Agents, Anthropic tool_runner, Claude Agent SDK}: 14/14 answered from the targeted document; every tool call carried the exact `doc_name`, no browse detours.
- Folders: 437/437 in test_client / test_agent_tools / test_local_chat with the folder tests; pyright: zero new diagnostics against the base; live: managed `chat(q, folder_id=F)` lists the folder's documents, `folder_id="root"` answers normally, own-model over cloud tools calls `browse_documents(folder_id=F, recursive=true)` first and gives the same list.

## Follow-ups (not in this PR)

- docs site `sdk/agents.mdx`: drop `doc_id` from the `agent_instructions()` / tool-helper examples and tables, document `document_context()`; land with the release. The folder docs (chat.mdx, documents.mdx) are in VectifyAI/pageindex-docs#53.
- Release notes: BREAKING line for the eight methods; `get_document()` now returns `metadata` locally too. Plus `folder_context()` and `chat(folder_id=)`.
- Cloud: `metadata` mixes internal keys (`thumbnailKey`, …) into user tags; compute's managed chat dumps `next_steps` / `user_id` into its own block; compute's `json.dumps` escapes non-ASCII names while the web app and SDK keep them raw.

https://claude.ai/code/session_01VxguPoTv2BmzS9d3erzTrS
https://claude.ai/code/session_01NsQjN5HMmmX4n5NVsZWYaw

### DIFF
--- a/examples/agentic_vectorless_rag_demo.py
+++ b/examples/agentic_vectorless_rag_demo.py
@@ -50,10 +50,11 @@ def query_agent(client: PageIndexLocalClient, doc_id: str, prompt: str, verbose:
     """
     agent = Agent(
         **client.openai_agent_config(
-            doc_id=doc_id,
             # model_settings=ModelSettings(reasoning={"effort": "low", "summary": "auto"}),  # from agents.model_settings import ModelSettings
         ),
     )
+    # Document targeting is conversation content: it leads the first message.
+    prompt = client.document_context(doc_id) + "\n\n" + prompt
 
     async def _run():
         streamed_run = Runner.run_streamed(agent, prompt)
--- a/pageindex/agent_tools.py
+++ b/pageindex/agent_tools.py
@@ -1744,3 +1744,45 @@ def doc_targeting_block(client, doc_id) -> Optional[str]:
         "Use these documents' names to retrieve their content with "
         "get_document_structure() and get_page_content()."
     )
+
+
+def folder_targeting_block(client, folder_id) -> Optional[str]:
+    """The folder_id targeting text, rendered as the cloud's managed chat
+    renders its own: the folder's name and metadata and the directive to
+    discover its documents there. None for no folder — None, "", and
+    "root", the library itself, which the managed chat leaves untargeted.
+    A folder proper is cloud-only: local libraries have none."""
+    if folder_id is None:
+        return None
+    if not isinstance(folder_id, str):
+        raise PageIndexAPIError("folder_id must be a string.")
+    if folder_id in ("", "root"):
+        return None
+    if not getattr(client, "api_key", None):
+        raise PageIndexAPIError(
+            "folder_id is cloud-only — folders are not supported in local "
+            "mode. Create the client with an api_key to use folders.")
+    folders = client.list_folders().get("folders") or []
+    folder = next((f for f in folders if f.get("id") == folder_id), None)
+    if folder is None:
+        raise PageIndexAPIError(
+            f"Folder not found or access denied: {folder_id}")
+    metadata = {key: folder[key] for key in ("id", "name", "description")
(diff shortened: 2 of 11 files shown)


rejojer:
@codex review

PR #496, opened 2026-09-09
chat(citations=True): cite answers with the server's cited_answer prompt

rejojer:
`chat(citations=True)` makes the answer cite every claim the way PageIndex chat does: `<cite doc="…" page="…"/>` tags, `block="…"` added where the cloud document has blocks. Stacks on #494 (bridge `prompts/get`); this PR's diff is the four files on top of it.

**How it works**
- Own-model chat over cloud documents: fetches the MCP server's `cited_answer` prompt (`format=cite`, the rules `citation-contract.ts` also feeds PageIndex chat) over the same bridge session as the tools, and folds it into the system prompt after the managed prompt and before the caller's `instructions`. All three lanes (answer lane, `protocol="responses"`, `protocol="messages"`) via the existing instructions plumbing; `local_chat.py` untouched.
- Managed chat: `citations=True` passes `enable_citations=True` to the endpoint.
- Local documents: a frozen copy of the same prompt (`LOCAL_CITATION_PROMPTS`, generated from the live text minus the one bullet naming the cloud-only `get_document_image()` tool). Local page content carries no blocks, so citations resolve to pages.
- `client.citation_prompt(format="cite")` exposes the same text for framework-built agents; `markdown` / `footnote` are the MCP prompt's variants for hosts that strip tags.

**Verified**
- 503 tests green (10 new, red-verified first), without-frameworks leg simulated, pyright clean.
- Key-gated live tests: the three formats served and distinct; local frozen copies equal the live text minus that bullet.
- Live with real models: block document → `<cite doc="document_5.pdf" page="1" block="p1_text_1"/>`; legacy document → `<cite doc="1706.03762_24.pdf" page="1"/>`; locally indexed PDF → `<cite doc="q1-fy25-earnings.pdf" page="3"/>`; managed `citations=True` non-stream and stream.

**Not in this PR**
- Docs (`agents.mdx`, `chat.mdx`'s hand-written `<cite>` system message becomes `citations=`) and a release-notes line, with the release.
- Server-side, for the prompt's owner: the managed API's citation prompt (pageindex-compute) carries a literal example id `p1_text_001` that the model echoes on legacy documents (resolved citations then carry a `block_id` with no `bbox`); and a `granularity` argument on `cited_answer` would let the SDK offer page-only citations on block documents.

https://claude.ai/code/session_01Ex6uKLsDrQjZkqypjSyPwA

### DIFF
--- a/pageindex/agent_tools.py
+++ b/pageindex/agent_tools.py
@@ -1649,19 +1649,19 @@ CITATIONS
 }
 
 
-def fetch_citation_prompt(client, format: Optional[str] = None) -> str:
+def fetch_citation_prompt(client, format: str) -> str:
     """The MCP server's ``cited_answer`` prompt as system-prompt text;
-    ``format`` rides as its one argument (None: the server's default).
-    Local: the frozen copy, page-level."""
+    ``format`` rides as its one argument. Local: the frozen copy,
+    page-level."""
     if not getattr(client, "api_key", None):
         try:
-            return LOCAL_CITATION_PROMPTS[format or "markdown"]
+            return LOCAL_CITATION_PROMPTS[format]
         except KeyError:
             raise PageIndexAPIError(
                 f"citations format {format!r} is not one of "
                 f"{', '.join(LOCAL_CITATION_PROMPTS)}.") from None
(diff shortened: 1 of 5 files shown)


rejojer:
@codex review

PR #497, opened 2026-09-09
Client instructions: a standing persona for every answer surface

rejojer:
## What

`PageIndexClient(..., instructions="...")` — standing guidance for the answering agent (persona, language, format), appended after the managed system prompt on every surface that produces an answer.

```python
client = PageIndexClient(
    index="cloud",
    chat="gpt-5.6-sol",
    instructions="You are ACME's support assistant. Answer in the user's language.",
)
client.chat("What is the refund policy?")                       # carries the persona
client.chat(history, instructions="This user prefers French.")  # persona + this call's addition
```

## Design

- **Name** `instructions`, the same word as `chat(instructions=)`. A client-level argument like `mode=`, not a chat-side spelling: it combines with a string `chat=` slot and never selects own-model chat on its own (a `chat_`-prefixed flat argument would have flipped a managed client to own-model chat through the config.yaml default).
- **Append only, never replace.** Prompt order: managed base → client `instructions` → `chat(instructions=)` → history `system` rows. Stable-before-variable is also the cache order. `""`/whitespace = unset, as `chat(instructions="")`.
- **Coverage.** One line in `_base_instructions` reaches every own-model surface: `chat()` / `chat_completions()`, `chat(protocol="responses")`, `chat(protocol="messages")` (inside the cached managed block), `agent_instructions()`, `openai_agent_config()`, `anthropic_runner_config()`, `claude_agent_config()`. Not touched: indexing, the tool surfaces, MCP tool calls, the cloud MCP server's own `initialize.instructions`.
- **Managed cloud chat.** The endpoint takes exactly one system message, first (it appends it to its own prompt as custom instructions). The client's instructions, the call's, and the history's system rows now fold into that one row, in that order. Two relaxations follow: `chat(instructions=)` works on a managed client (refused in the SDK since #460, although the endpoint has accepted custom instructions since Sep 1), and a `system` row anywhere in the history no longer 400s.
- **Thin fold.** The SDK lifts the system/developer rows into that one leading row (blank ones dropped) and sends the rest of the history to the endpoint as given, every field included. What the SDK reads, it validates: a system row whose content is not text (or a list of text parts) is refused here. The rest the endpoint decides on (it refuses tool rows and structured user content itself: 400 / 422, live-verified).

## Verification

- 13 new tests, red-verified: constructor storage/blank/type on all three classes; append on local and on the live-MCP bridge; prompt order on the answer, Responses and Messages lanes; an end-to-end `chat()` call on the own-model lane (the persona and the call's system text reach the model; red-verified by blanking the persona on that lane alone); the three config bundles; the managed fold (order, streamed door, developer rows) and the verbatim forwarding of the rest of the history (a leading system row kept in place, blank system rows dropped, tool rows / structured content / extra fields untouched). `test_cloud_guards` drops its pin on the old managed refusal.
- 501 passed; pyright 0 errors on the edited files; the without-frameworks CI leg simulated (agents / anthropic / claude_agent_sdk blocked): passes with the framework tests skipped.
- Live: a managed client with `instructions="…French…PINEAPPLE"` answers in French; with a per-call addition both layers show; a bridge client's `agent_instructions()` ends with the persona after the live MCP guidance.

## Pending (land with the release)

- Docs: `client.mdx` constructor table row, `chat.mdx` client example + `instructions` parameter row, `agents.mdx` one sentence on `agent_instructions()`.
- Release notes: an "Also in" line for the new argument and the two managed-lane relaxations.

https://claude.ai/code/session_01J8fbpdM5pz2JLjiNY11usy
https://claude.ai/code/session_01TgdXZx63aMwrpaKFTQFFKx

## Rebased onto 0.2.16's other lanes

Two test pins this branch voids on purpose: the chat_completions protocol lane's guard pinned the managed refusal of `instructions=`, which the fold lifts, and `_anthropic_system` lost its `doc_id` argument when document targeting moved to the first user message.

### DIFF
--- a/pageindex/agent_tools.py
+++ b/pageindex/agent_tools.py
@@ -1691,16 +1691,19 @@ def _base_instructions(client, include_management: bool = False) -> str:
     """Cloud: the live instructions the MCP server serves for the tool set
     actually shipped. Local: the built-in subset instructions."""
     if not getattr(client, "api_key", None):
-        return AGENT_INSTRUCTIONS
-    instructions = _cloud_bridge(
-        client, gated=not include_management).instructions()
-    if not isinstance(instructions, str) or not instructions.strip():
-        raise PageIndexAPIError(
-            "The MCP server returned no agent instructions — refusing to "
-            "substitute the SDK's local-subset guidance, which does not "
-            "cover the cloud tool set."
-        )
-    return instructions
+        base = AGENT_INSTRUCTIONS
+    else:
+        base = _cloud_bridge(
+            client, gated=not include_management).instructions()
+        if not isinstance(base, str) or not base.strip():
+            raise PageIndexAPIError(
+                "The MCP server returned no agent instructions — refusing "
+                "to substitute the SDK's local-subset guidance, which does "
+                "not cover the cloud tool set."
+            )
+    # The client's standing instructions follow the base on every surface.
+    own = getattr(client, "instructions", None)
+    return f"{base}\n\n{own}" if own else base
 
 
 def doc_targeting_block(client, doc_id) -> Optional[str]:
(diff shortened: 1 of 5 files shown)


rejojer:
@codex review

